PR #3, opened 2019-08-13
Extend, fix & add docs with sphinx-argparse

rohanpm:
Initial version of "command reference" docs.
Each command should be documented on its own page.

This documentation can be (partially) generated from the
ArgumentParser objects, to document all supported arguments with
their help text.

Some minor refactoring was needed to ensure that ArgumentParser
objects can be created & accessed without running tasks.
Most notably, the usage of attrs was dropped as it mainly
seems to be getting in the way, and we don't much benefit from
any attrs features here.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,6 +45,7 @@ extensions = [
     "sphinx.ext.napoleon",
     "sphinx.ext.githubpages",
     "sphinx.ext.viewcode",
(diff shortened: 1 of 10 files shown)


rohanpm:
Looks like this isn't passing because coveralls is currently in maintenance, but if you dig into the logs you can see it's all working with 100% coverage.

PR #6, opened 2019-08-13
code refactor on review

rajulkumar:
- rename pulp to _pulp to reflect no public API in pubtools._pulp
- remove mocking PulpTask add_args in tests
- updated parser options url, user and password to include 'pulp'


rohanpm:
What's here looks fine, but I think it will break docs now due to _pulp rename without corresponding rename in garbage-collect.rst, could you have a look?

rajulkumar:
updated garbage-collect.rst as mentioned above

PR #7, opened 2019-08-14
updated tests to remove using internal apis from pubtools-pulplib

### LINKED ISSUE #5: Tests use private API
#### Summary

Autotests in this project currently use private API from pubtools-pulplib.

https://github.com/release-engineering/pubtools-pulp/blob/master/tests/test_garbage_collect.py, note lines including "import pubtools.pulplib._impl".

#### Issue Type
- Bug Report

#### Steps to reproduce
- Review code

#### Actual results
- Review finds import of private modules (modules with name beginning with '_')

#### Expected results
- Review doesn't find import of private modules 

#### OS / Environment
all

#### Additional Information

https://github.com/release-engineering/pubtools-pulplib/pull/35 is a PR to help remove the need for this usage of private API.

rajulkumar:
fix #5

### DIFF
--- a/tests/test_garbage_collect.py
+++ b/tests/test_garbage_collect.py
@@ -1,5 +1,4 @@
 import pytest
-import attr
 import datetime
 from mock import Mock, patch
 from more_executors.futures import f_return
@@ -11,8 +10,6 @@ from pubtools.pulplib import (
     Task,
     InvalidDataException,
 )
-import pubtools.pulplib._impl.model.attr as pp_attr
-import pubtools.pulplib._impl.compat_attr as attr
 
 from pubtools.pulp.tasks.garbage_collect import GarbageCollect, entry_point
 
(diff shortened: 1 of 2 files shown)


tests/test_garbage_collect.py
```diff
@@ -30,15 +27,9 @@ def _get_time_created(d=0, h=0, s=0):
...
-    @attr.s(kw_only=True, frozen=True)
-    class TestRepo(Repository):
-        is_temp_repo = pp_attr.pulp_attrib(
-            default=None, type=str, pulp_field="notes.pub_temp_repo"
-        )
-
     controller = FakeController()
     for repo in args:
-        yum_repo = TestRepo.from_data(repo)
+        yum_repo = Repository.from_data(repo)
```
rohanpm: Really I don't think you should need to use from_data at all.  Can you just create the repository objects with the attributes you need?  It's intended to be used that way, so converting from raw Pulp data happens in a bare minimum of places.

You would probably have to help me get https://github.com/release-engineering/pubtools-pulplib/pull/36 done and released first though.

rajulkumar: Updated the tests to use created repo objects using Repository attributes

tests/test_garbage_collect.py
```diff
@@ -23,23 +20,14 @@ def mock_logger():
         yield mocked_info
 
 
-def _get_time_created(d=0, h=0, s=0):
-    return (
-        datetime.datetime.utcnow() - datetime.timedelta(days=d, hours=h, seconds=s)
-    ).strftime("%Y-%m-%dT%H:%M:%SZ")
+def _get_created(d=0, h=0, s=0):
```
rohanpm: Minor: I don't think you need to be naming these things with a leading `_` (up to you though).

rajulkumar: It's only to identify helper functions. Fine with changing this convention.

PR #9, opened 2019-09-03
Introduce clear-repo command

rohanpm:
Implements the clear-repo workflow, which is:

- delete all content of selected type(s) from selected Pulp repo(s)
- record info on deleted push items
- publish the repos
- do UD cache flushing
- do CDN cache flushing

A few refactors were made and (internal) features added to support
this new task:

- `@step` decorator applies standard plaintext and machine-readable
  logs around discrete workflow steps.  Supports blocking and
  non-blocking coding styles.

- Service classes were introduced to encapsulate the creation of
  a particular service based on command-line arguments. Tasks should
  inherit from the Services they want to use. For example, if a task
  needs to use fast purge, it should inherit FastPurgeClientService
  to add all needed command line arguments and properties.

- Simple UD cache flushing client was added

- CommandTester helper class was added to make it easy to write
  log-based tests (see test/logs included in this commit).

### DIFF
--- a/pubtools/_pulp/services/pulp.py
+++ b/pubtools/_pulp/services/pulp.py
@@ -1,0 +1,69 @@
+import threading
+import os
+import logging
+import warnings
+
+from pubtools import pulplib
+
+from .base import Service
+
+LOG = logging.getLogger("pubtools.pulp")
+
+
+class PulpClientService(Service):
+    """A service providing a Pulp client.
+
+    If this service is inherited, Pulp-related arguments become mandatory
+    in order to run the task.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.__lock = threading.Lock()
+        self.__instance = None
+        super(PulpClientService, self).__init__(*args, **kwargs)
+
+    def add_service_args(self, parser):
+        super(PulpClientService, self).add_service_args(parser)
+
+        group = parser.add_argument_group("Pulp environment")
+        group.add_argument("--pulp-url", help="Pulp server URL", required=True)
+        group.add_argument("--pulp-user", help="Pulp username", default=None)
+        group.add_argument(
+            "--pulp-password",
+            help="Pulp password (or set PULP_PASSWORD environment variable)",
+            default=None,
+        )
+        group.add_argument(
+            "--pulp-insecure",
+            action="store_true",
+            help="Allow unverified HTTPS connection to Pulp",
+        )
+
+    @property
+    def pulp_client(self):
+        """A Pulp client used during task, instantiated on demand."""
+        with self.__lock:
+            if not self.__instance:
+                self.__instance = self.__get_instance()
+        return self.__instance
+
+    def __get_instance(self):
+        auth = None
+        args = self._service_args
+
+        # checks if pulp password is available as environment variable
+        if args.pulp_user:
+            pulp_password = args.pulp_password or os.environ.get("PULP_PASSWORD")
+            if not pulp_password:
+                LOG.warning("No pulp password provided for %s", args.pulp_user)
+            auth = (args.pulp_user, pulp_password)
+
+        kwargs = {"auth": auth}
+
+        if args.pulp_insecure:
+            kwargs["verify"] = False
+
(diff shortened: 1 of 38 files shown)


pubtools/_pulp/step.py
```diff
@@ -0,0 +1,137 @@
...
+                return args[0] if args else None
+
+            logger = StepLogger(self)
+            logger.log_start(args)
+
+            try:
+                ret = fn(instance, *args, **kwargs)
+            except SystemExit as exc:
+                if exc.code == 0:
+                    logger.log_return([None])
```
rohanpm:
```suggestion
                    logger.log_return()
```

rohanpm:
@JayZ12138, @rajulkumar as I think you're both working on adding new tasks to this project, could you please prioritize review of this?

Your work may be affected by the refactors mentioned in the PR description.

tests/clear_repo/test_clear_repo.py
```diff
@@ -0,0 +1,326 @@
...
+
+    @property
+    def fastpurge_client(self):
+        # Super may or may not give a fastpurge client, depends on arguments
+        from_super = super(FakeClearRepo, self).fastpurge_client
+        if from_super:
+            # If it did create one, it should be this
+            assert isinstance(from_super, FastPurgeClient)
+
+        # We'll substitute our own, only if UD client is being used
```
rohanpm:
```suggestion
        # We'll substitute our own, only if fastpurge client is being used
```

pubtools/_pulp/services/pulp.py
```diff
@@ -0,0 +1,69 @@
...
+            action="store_true",
+            help="Allow unverified HTTPS connection to Pulp",
+        )
+
+    @property
+    def pulp_client(self):
+        """A Pulp client used during task, instantiated on demand."""
+        with self.__lock:
+            if not self.__instance:
+                self.__instance = self.__get_instance()
```
JayZ12138: I don't understand this. 

looks like __get_instance would only be called when there is no __instance bound, but the clearRepo task inherits 4 services and it gets all kinds of clients, I don't know why:
1. isn't __get_instance overwritten by the last parent? 
2. __get_instance called once and it assigns the instance to self._instance and it shouldn't be called again, why it get all 4 service clients?

rohanpm: I think what you're missing is the special behavior of `__` prefix:

In Python, if you name a class property or method with leading `__`, it's mangled to include the class name.  That's specifically so that you can reuse the same name in multiple classes within a hierarchy, without clashes.

See this demo:

```
$ python3
Python 3.7.4 (default, Jul  9 2019, 16:32:37) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> class Klass:
...   __my_variable = 'abc'
... 

# In dir(), we can see the real name used to store it is _Klass__my_variable
>>> dir(Klass)
['_Klass__my_variable', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__']

# From outside the class, I can't access it with the declared name
>>> Klass.__my_variable
  File "<stdin>", line 1, in <module>
AttributeError: type object 'Klass' has no attribute '__my_variable'

# Because it's actually named this
>>> Klass._Klass__my_variable
'abc'
```

It's good practice to use this on classes designed to be inherited, because otherwise everyone inheriting your class has to know what names you've used, so they can avoid clashes.

JayZ12138: ah, yeah, I completely forget this feature of python. 
I just remember prefix with double-underscore is the recommended way to make objects private in python but nobody uses it...
Never expected it could achieve such things.

PR #10, opened 2019-09-09
Introduce set maintenance commands

JayZ12138:
Two commands are included:
 1. pubtools-pulp-maintenance-on: enable maintenance mode
 2. pubtools-pulp-maintenance-off: disale maintenance mode

Both commands accepts repo ids and repo regex pattern as input. For enable
maintenance mode, if repo ids are passed, then it first checks if the repo
exists or not by querying pulp server with repo ids.

The general workflow for set maintenance is:
 1. get maintenance report from pulp server
 2. make changes to report according to enable/disable
 3. upload result report to pulp and publish

### DIFF
--- a/docs/set-maintenance.rst
+++ b/docs/set-maintenance.rst
@@ -20,7 +20,7 @@ A typical invocation of set maintenance would look like this:
     --pulp-url https://pulp.example.com/ \
     --pulp-user admin \
     --pulp-password XXXXX \
-    --repo-id example-repo
+    --repo-ids example-repo
 
 In this example, the repository with id 'example-repo' will be set to
 maintenance mode.
@@ -45,7 +45,7 @@ A typical invocation of unset maintenance would look like this:
     --pulp-url https://pulp.example.com/ \
     --pulp-user admin \
     --pulp-password XXXXX \
-    --repo-id example-repo
+    --repo-ids example-repo
 
 In this example, the repository with id 'example-repo' will be removed
 from maintenance mode.
--- a/docs/set-maintenance.rst
+++ b/docs/set-maintenance.rst
@@ -1,0 +1,51 @@
+set-maintenance
+===============
+
+Set Maintenance On
+-------------------
+
+.. argparse::
+   :module: pubtools._pulp.tasks.set_maintenance.set_maintenance_on
+   :func: doc_parser
+   :prog: pubtools-pulp-maintenance-on
+
+Example
+.......
+
+A typical invocation of set maintenance would look like this:
+
+.. code-block::
+
+  pubtools-pulp-maintenance-on \
+    --pulp-url https://pulp.example.com/ \
+    --pulp-user admin \
+    --pulp-password XXXXX \
+    --repo-id example-repo
+
+In this example, the repository with id 'example-repo' will be set to
+maintenance mode.
+
+Set Maintenance Off
+-------------------
+
+.. argparse::
+   :module: pubtools._pulp.tasks.set_maintenance.set_maintenance_off
+   :func: doc_parser
+   :prog: pubtools-pulp-maintenance-off
+
+
+Example
+.......
+
+A typical invocation of unset maintenance would look like this:
+
+.. code-block::
+
+  pubtools-pulp-maintenance-off \
+    --pulp-url https://pulp.example.com/ \
+    --pulp-user admin \
+    --pulp-password XXXXX \
+    --repo-id example-repo
+
+In this example, the repository with id 'example-repo' will be removed
+from maintenance mode.
--- a/pubtools/_pulp/tasks/set_maintenance/base.py
+++ b/pubtools/_pulp/tasks/set_maintenance/base.py
@@ -15,8 +15,8 @@ class SetMaintenance(PulpClientService, PulpTask):
         self.parser.add_argument("--owner", help="who sets/unsets maintenance mode")
 
         self.parser.add_argument(
-            "--repo-regex",
-            help="only set repositories matched this regex to maintenance mode",
+            "--repo-url-regex",
+            help="set repositories whose relative_url matches to this regex to maintenance mode",
         )
 
         self.parser.add_argument(
--- a/pubtools/_pulp/tasks/set_maintenance/base.py
+++ b/pubtools/_pulp/tasks/set_maintenance/base.py
@@ -1,4 +1,5 @@
 import logging
+import re
 
 from pubtools._pulp.task import PulpTask
 from pubtools._pulp.services import PulpClientService
@@ -16,7 +17,8 @@ class SetMaintenance(PulpClientService, PulpTask):
 
         self.parser.add_argument(
             "--repo-url-regex",
-            help="set repositories whose relative_url matches to this regex to maintenance mode",
+            help="adjust maintenance mode for repositories with a publish URL matching this pattern",
+            type=re.compile,
         )
 
         self.parser.add_argument(
@@ -42,6 +44,4 @@ class SetMaintenance(PulpClientService, PulpTask):
 
         report = self.adjust_maintenance_report(report)
 
-        tasks = self.set_maintenance(report).result()
-        task_ids = ", ".join([t.id for t in tasks])
-        LOG.info("Completed adjusting maintenance mode: %s", task_ids)
+        self.set_maintenance(report).result()
--- a/pubtools/_pulp/tasks/set_maintenance/base.py
+++ b/pubtools/_pulp/tasks/set_maintenance/base.py
@@ -1,0 +1,20 @@
+from pubtools._pulp.task import PulpTask
+from pubtools._pulp.services import PulpClientService
+
+
+class SetMaintenance(PulpClientService, PulpTask):
+    def add_args(self):
+        super(SetMaintenance, self).add_args()
+
+        self.parser.add_argument("--owner", help="who sets/unsets maintenance mode")
+
+        self.parser.add_argument(
+            "--repo-regex",
+            help="only set repositories matched this regex to maintenance mode",
+        )
+
+        self.parser.add_argument(
+            "--repo-ids",
+            nargs="+",
+            help="repository to be set/unset to maintenance mode",
+        )
--- a/pubtools/_pulp/tasks/set_maintenance/base.py
+++ b/pubtools/_pulp/tasks/set_maintenance/base.py
@@ -1,5 +1,11 @@
+import logging
+
 from pubtools._pulp.task import PulpTask
 from pubtools._pulp.services import PulpClientService
+
+
+step = PulpTask.step
+LOG = logging.getLogger("set-maintenance")
 
 
 class SetMaintenance(PulpClientService, PulpTask):
@@ -18,3 +24,24 @@ class SetMaintenance(PulpClientService, PulpTask):
             nargs="+",
             help="repository to be set/unset to maintenance mode",
         )
+
+    @step("Get maintenance report")
+    def get_maintenance_report(self):
+        return self.pulp_client.get_maintenance_report()
+
+    @step("Adjust maintenance report")
+    def adjust_maintenance_report(self, report):
+        raise NotImplementedError
+
+    @step("Set maintenance report")
+    def set_maintenance(self, report):
+        return self.pulp_client.set_maintenance(report)
+
+    def run(self):
+        report = self.get_maintenance_report().result()
+
+        report = self.adjust_maintenance_report(report)
+
+        tasks = self.set_maintenance(report).result()
+        task_ids = ", ".join([t.id for t in tasks])
+        LOG.info("Completed adjusting maintenance mode: %s", task_ids)
--- a/pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
+++ b/pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
@@ -1,0 +1,48 @@
+import logging
+import re
+
+from .base import SetMaintenance
+
+
+LOG = logging.getLogger("set-maintenance")
+
+
+class SetMaintenanceOff(SetMaintenance):
+    """Unset repositories maintenance mode.
+
+    See "pub maintenance-on --help" for more information on maintenance mode.
+    """
+
+    def run(self):
+        report = self.pulp_client.get_maintenance_report().result()
+
+        to_remove = []
+        if self.args.repo_regex:
+            for entry in report.entries:
+                if re.match(self.args.repo_regex, entry.repo_id):
+                    to_remove.append(entry.repo_id)
+        if self.args.repo_ids:
+            existed_repo_ids = [entry.repo_id for entry in report.entries]
+            for repo_id in self.args.repo_ids:
+
+                if repo_id in existed_repo_ids:
+                    to_remove.append(repo_id)
+                else:
+                    LOG.warn("Repository %s is not in Maintenance Mode", repo_id)
+
+        LOG.info(
+            "Following repositories will be removed from Maintenance Mode: \n%s",
+            "\n".join(to_remove),
+        )
+
+        report = report.remove(to_remove, owner=self.args.owner)
+
+        self.pulp_client.set_maintenance(report).result()
+
+
+def entry_point():
+    SetMaintenanceOff().main()
+
+
+def doc_parser():
+    return SetMaintenanceOff().parser
--- a/pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
+++ b/pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
@@ -2,20 +2,18 @@ import logging
 import re
 
 from .base import SetMaintenance
+from pubtools._pulp.task import PulpTask
 
 
+step = PulpTask.step
 LOG = logging.getLogger("set-maintenance")
 
 
 class SetMaintenanceOff(SetMaintenance):
-    """Unset repositories maintenance mode.
+    """Unset repositories maintenance mode."""
 
-    See "pub maintenance-on --help" for more information on maintenance mode.
-    """
-
-    def run(self):
-        report = self.pulp_client.get_maintenance_report().result()
-
+    @step("Adjust maintenance report")
+    def adjust_maintenance_report(self, report):
         to_remove = []
         if self.args.repo_regex:
             for entry in report.entries:
@@ -28,20 +26,18 @@ class SetMaintenanceOff(SetMaintenance):
                 if repo_id in existed_repo_ids:
                     to_remove.append(repo_id)
                 else:
-                    LOG.warn("Repository %s is not in Maintenance Mode", repo_id)
-
+                    LOG.warning("Repository %s is not in maintenance mode", repo_id)
         LOG.info(
-            "Following repositories will be removed from Maintenance Mode: \n%s",
+            "Following repositories will be removed from maintenance mode: \n%s",
             "\n".join(to_remove),
         )
-
         report = report.remove(to_remove, owner=self.args.owner)
 
-        self.pulp_client.set_maintenance(report).result()
+        return report
 
 
-def entry_point():
-    SetMaintenanceOff().main()
+def entry_point(cls=SetMaintenanceOff):
+    cls().main()  # pragma: no cover
 
 
 def doc_parser():
--- a/pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
+++ b/pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
@@ -1,6 +1,8 @@
 import logging
 import re
 
+from pubtools.pulplib import Criteria
+
 from .base import SetMaintenance
 from pubtools._pulp.task import PulpTask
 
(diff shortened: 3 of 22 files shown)


docs/set-maintenance.rst
```diff
@@ -0,0 +1,51 @@
...
+
+A typical invocation of set maintenance would look like this:
+
+.. code-block::
+
+  pubtools-pulp-maintenance-on \
+    --pulp-url https://pulp.example.com/ \
+    --pulp-user admin \
+    --pulp-password XXXXX \
+    --repo-id example-repo
```
rohanpm: Docs don't match implementation (your code says --repo-ids).

pubtools/_pulp/tasks/set_maintenance/set_maintenance_on.py
```diff
@@ -0,0 +1,66 @@
...
+
+        self.pulp_client.set_maintenance(report).result()
+
+    def _ensure_repos_exist(self, repo_ids):
+        """Check if repositories are existed in Pulp server, if not, PulpException will
+        be raised.
+        """
+        repo_ft = []
+        for repo in repo_ids:
+            repo_ft.append(self.pulp_client.get_repository(repo))
```
rohanpm: search_repository could end up being a fair bit faster here, if there's hundreds of repos (possible).  That'll do one request to Pulp per page (~2000 repos) instead of one request to Pulp per repo.

```
   found_repos = self.pulp_client.search_repository(Criteria.with_ids(repo_ids)).result()
   found_ids = [r.id for r in found_repos]
   missing_ids = set(repo_ids) - set(found_ids)
   if missing_ids:
      # ... raise the error here
```

tests/set_maintenance/test_set_maintenance.py
```diff
@@ -0,0 +1,178 @@
...
+            s_m.main()
+    return controller
+
+
+def test_maintenance_on(mock_on_logger):
+    """Test set maintenance by passing repo ids."""
+    repo1 = Repository(id="repo1")
+    repo2 = Repository(id="repo2")
+
+    cmd_args = ["--repo-id", "repo1", "repo2", "--message", "Now in Miantenance"]
```
rohanpm:
```suggestion
    cmd_args = ["--repo-id", "repo1", "repo2", "--message", "Now in Maintenance"]
```

tests/set_maintenance/test_set_maintenance.py
```diff
@@ -0,0 +1,178 @@
...
+
+    controller = _run_test(cmd_args, True, repo1, repo2)
+
+    # upload and publish should be called once each
+    assert len(controller.upload_history) == 1
+    assert controller.upload_history[0].repository.id == "redhat-maintenance"
+    assert len(controller.publish_history) == 1
+    assert controller.publish_history[0].repository.id == "redhat-maintenance"
+
+    # logged message should indicate both repos are set to maintenance
```
rohanpm: 1. Instead of testing logged messages manually, can you please consider using the command_tester fixture which would test the entire log at once?  This would protect against e.g. completely bogus log messages left in the code accidentally, and also let code reviewers see how the entire task is expected to look in practice.

2. If you don't want to do that, instead of mocking the logger, can you please consider using caplog fixture instead?  This would at least let the test keep working without depending on that specific logger and method call being used.

tests/set_maintenance/test_set_maintenance.py
```diff
@@ -0,0 +1,178 @@
...
+    args = ["", "--pulp-url", "http://some.url", "--verbose", "--repo-regex", "repo1"]
+    s_m = SetMaintenanceOff()
+
+    # remove repos from maintenance mode by id
+    with patch("sys.argv", args):
+        with _patch_pulp_client(controller.client):
+            s_m.main()
+
+    mock_off_logger.info.assert_called_with(
+        "Following repositories will be removed from Maintenance Mode: \n%s", "repo1"
```
rohanpm: I think at least one of these tests (or ideally even all) should get the maintenance report from the fake client and check that it matches the expected outcome, so it's not only testing that it *claimed to* adjust maintenance mode.

pubtools/_pulp/tasks/set_maintenance/set_maintenance_on.py
```diff
@@ -0,0 +1,66 @@
...
+        report = self.pulp_client.get_maintenance_report().result()
+        repo_ids = []
+        if self.args.repo_ids:
+            self._ensure_repos_exist(self.args.repo_ids)
+            repo_ids.extend(self.args.repo_ids)
+
+        if self.args.repo_regex:
+            crit = Criteria.with_field("id", Matcher.regex(self.args.repo_regex))
+            repos = self.pulp_client.search_repository(crit).result()
+            repo_ids.extend([repo.id for repo in repos.as_iter()])
```
rohanpm: No need for as_iter now.

```suggestion
            repo_ids.extend([repo.id for repo in repos])
```

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
```diff
@@ -0,0 +1,48 @@
...
+from .base import SetMaintenance
+
+
+LOG = logging.getLogger("set-maintenance")
+
+
+class SetMaintenanceOff(SetMaintenance):
+    """Unset repositories maintenance mode.
+
+    See "pub maintenance-on --help" for more information on maintenance mode.
```
rohanpm: I would rather drop the specific mention of Pub.  It should be possible to understand what this command is doing without having to refer back to Pub.

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
```diff
@@ -0,0 +1,48 @@
...
+                if re.match(self.args.repo_regex, entry.repo_id):
+                    to_remove.append(entry.repo_id)
+        if self.args.repo_ids:
+            existed_repo_ids = [entry.repo_id for entry in report.entries]
+            for repo_id in self.args.repo_ids:
+
+                if repo_id in existed_repo_ids:
+                    to_remove.append(repo_id)
+                else:
+                    LOG.warn("Repository %s is not in Maintenance Mode", repo_id)
```
rohanpm: I don't think we need Maintenance Mode capitalized like this.

```suggestion
                    LOG.warn("Repository %s is not in maintenance mode", repo_id)
```

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
```diff
@@ -0,0 +1,48 @@
...
+
+        LOG.info(
+            "Following repositories will be removed from Maintenance Mode: \n%s",
```
rohanpm:
```suggestion
            "Following repositories will be removed from maintenance mode: \n%s",
```

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
```diff
@@ -0,0 +1,48 @@
...
+    """Unset repositories maintenance mode.
+
+    See "pub maintenance-on --help" for more information on maintenance mode.
+    """
+
+    def run(self):
+        report = self.pulp_client.get_maintenance_report().result()
+
+        to_remove = []
+        if self.args.repo_regex:
```
rohanpm: I'd like to suggest putting all this "get repos" logic into a separate method and marking it with a ``@step`` decorator.  Getting/adjusting/uploading the maintenance mode possibly should be a ``@step`` as well.

I think we should try to use steps for just about every task, because it's a prerequisite for tasks to display meaningfully in the planned progress UI using the machine-readable logs.

pubtools/_pulp/tasks/set_maintenance/set_maintenance_on.py
```diff
@@ -0,0 +1,66 @@
...
+
+        report = report.add(repo_ids, owner=self.args.owner, message=self.args.message)
+
+        LOG.info(
+            "Setting following repos to Maintenance Mode: \n%s", "\n".join(repo_ids)
```
rohanpm:
```suggestion
            "Setting following repos to maintenance mode: \n%s", "\n".join(repo_ids)
```

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
JayZ12138: Makes sense.

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
```diff
@@ -0,0 +1,48 @@
...
+                    LOG.warn("Repository %s is not in Maintenance Mode", repo_id)
+
+        LOG.info(
+            "Following repositories will be removed from Maintenance Mode: \n%s",
+            "\n".join(to_remove),
+        )
+
+        report = report.remove(to_remove, owner=self.args.owner)
+
+        self.pulp_client.set_maintenance(report).result()
```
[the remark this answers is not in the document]
JayZ12138: Yup, it returns some tasks. I tried locally and seems like task ids are logged by pubtools-pulplib already, do you think we should log them again here?

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
```diff
@@ -0,0 +1,48 @@
...
+    See "pub maintenance-on --help" for more information on maintenance mode.
+    """
+
+    def run(self):
+        report = self.pulp_client.get_maintenance_report().result()
+
+        to_remove = []
+        if self.args.repo_regex:
+            for entry in report.entries:
+                if re.match(self.args.repo_regex, entry.repo_id):
```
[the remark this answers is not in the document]
JayZ12138: yes, ran regex against the URL sounds more useful. 
The reason I use regex against id is because there is no URL in the maintenance report, every entry only contains id, message and a timestamp. For removing entries from the report, run regex against URL is not usable and then the user needs to pass all repo ids they want to remove from maintenance mode. 

I'm thinking I need to add URL to maintenance entry?

pubtools/_pulp/tasks/set_maintenance/set_maintenance_on.py
JayZ12138: Instead of raising, how about just removing the repo ids and log a warning? make it consistent with the behavior of set-maintenance-off when the entry is not in the report and more user-friendly

JayZ12138:
> Based on the CI failure, looks like pubtools-pulplib might need another bugfix too.

I checked, it's because there's nothing to do with the maintenance report, it just created a new report with last_updated defaulted to None and then error raised in write_timestamp. 

I'm thinking this could be a wrong use case and could rarely happen: 
1. it only happens when there's no report in the repo, which has to be the first time the maintenance mode used. 
2. it only happens it's first used and the user tries to unset maintenance mode. It's obviously wrong. 

Not sure if we should make a pull request to fix this?

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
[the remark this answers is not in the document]
rohanpm: I don't think you need to add URLs to the maintenance entries - the entry gives you the repo ID and that gives you enough info to find the repos, which then allows you to check their URLs.  Or do you have a concern with doing it that way?  It doesn't seem like there'd be a performance issue, it only costs one additional Pulp repo search.

tests/logs/set_maintenance/test_set_maintenance/test_maintenance_off.txt
```diff
@@ -0,0 +1,9 @@
+[    INFO] Get maintenance report: started
+[    INFO] Get maintenance report: finished
+[    INFO] Adjust maintenance report: started
+[    INFO] Following repositories will be removed from maintenance mode: 
+repo2
```
rohanpm: I think this is showing that it would be quite nice to format this somehow, e.g. indent it.

On the real production system, the prefix of each log message is going to be even deeper (with a timestamp as well), so this is going to look very disconnected, arguably quite ugly as the "Following repositories..." message is deeply indented but then each repo is not.

What do you think?  Maybe play around with either indenting this by a fair number of spaces, or else make each repo ID have its own call to the logger?

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
```diff
@@ -0,0 +1,44 @@
...
+            "Following repositories will be removed from maintenance mode: \n%s",
+            "\n".join(to_remove),
+        )
+        report = report.remove(to_remove, owner=self.args.owner)
+
+        return report
+
+
+def entry_point(cls=SetMaintenanceOff):
+    cls().main()  # pragma: no cover
```
rohanpm: In ClearRepo, the reason why I accepted the class in the entry point was exactly so that I didn't need to omit it from coverage - as I could have at least one test where I passed in my testing subclass of ClearRepo to the entry point.

Here, you've accepted a class as input but then you've also got pragma: no cover.  It looks like in your tests, you did indeed subclass the task classes, so could you also easily cover this entry point by passing it in from at least one test?  Would remove the need for this pragma.

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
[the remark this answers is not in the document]
rohanpm: I think it's important to have some hint in the logs about what a task did, otherwise it will look like a bunch of meaningless indistinguishable IDs.  But I agree it doesn't look good to have all the tasks logged twice in different ways.

A couple of ideas what to do here:

1. Do what I suggested here (and always do that in pubtools-pulp), and downgrade the related log in pulplib to DEBUG only.
2. Or, don't add any logging here.  Instead we can enhance the pulplib logging to also include the task tags.  Tags typically contain enough info to distinguish different tasks (e.g. the tags will include "publish" and the repo ID, when publishing a repo).

Maybe you could come up with some other ideas too if you play around with it.  At the moment I'm thinking option (2) ought to work out well.  If you think so too then I'd withdraw this suggestion and we don't need to log task IDs here.

pubtools/_pulp/tasks/set_maintenance/set_maintenance_on.py
rohanpm: Yeah, warning is better.

rohanpm:
> > Based on the CI failure, looks like pubtools-pulplib might need another bugfix too.
> 
> I checked, it's because there's nothing to do with the maintenance report, it just created a new report with last_updated defaulted to None and then error raised in write_timestamp.
> 
> I'm thinking this could be a wrong use case and could rarely happen:
> 
> 1. it only happens when there's no report in the repo, which has to be the first time the maintenance mode used.
> 2. it only happens it's first used and the user tries to unset maintenance mode. It's obviously wrong.
> 
> Not sure if we should make a pull request to fix this?

I count it as a bug in pulplib, because even if this use is invalid, I think it should be raising some kind of controlled error with a meaningful message (e.g. by an attrs validator) and not a crash by attempted None dereference.  But - I don't necessarily think that bug is too important, if there are other ways to get the tests passing here.

tests/logs/set_maintenance/test_set_maintenance/test_maintenance_off.txt
JayZ12138: yeah, I agree. 
I actually tried to use ' - ' as indent, but it didn't look better. I also tried to make each repo ID have its own call to the logger, but still didn't improve much... I guess I would need at least one more commit to finish this PR, so put it on hold then. Maybe combine them is the answer.

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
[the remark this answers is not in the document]
JayZ12138: ah, makes sense.

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
JayZ12138: I could see you must do this for test coverage, but didn't find out what the real magic is...

JayZ12138:
pubtools.pulplib doesn't support search by relative_url. Made a PR already. 

You could review the rest except the --repo-regex part if you want.

pubtools/_pulp/tasks/set_maintenance/base.py
```diff
@@ -0,0 +1,47 @@
...
+
+class SetMaintenance(PulpClientService, PulpTask):
+    def add_args(self):
+        super(SetMaintenance, self).add_args()
+
+        self.parser.add_argument("--owner", help="who sets/unsets maintenance mode")
+
+        self.parser.add_argument(
+            "--repo-url-regex",
+            help="set repositories whose relative_url matches to this regex to maintenance mode",
```
rohanpm: This is a base class but this help message only seems true for the SetMaintenanceOn case.  At least, my reading of this text is that it means maintenance mode will be *enabled* for the matched repos.

Also, as a minor thing I think it's better to have a user-oriented term like "publish URL" than the implementation detail of a field named relative_url.


Would this be better?

```suggestion
            help="adjust maintenance mode for repositories with a publish URL matching this pattern",
```

pubtools/_pulp/tasks/set_maintenance/base.py
```diff
@@ -0,0 +1,47 @@
...
+LOG = logging.getLogger("set-maintenance")
+
+
+class SetMaintenance(PulpClientService, PulpTask):
+    def add_args(self):
+        super(SetMaintenance, self).add_args()
+
+        self.parser.add_argument("--owner", help="who sets/unsets maintenance mode")
+
+        self.parser.add_argument(
```
rohanpm: You might want to consider adding a `type` here which compiles the regex, the main benefit being that the code will fail as early as possible if the user passes something which is not a valid regex.  Without this, I guess it won't fail until part way through the task.  A minor suggestion only.

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
```diff
@@ -0,0 +1,55 @@
...
+            for repo_id in to_remove:
+                LOG.info(" - %s", repo_id)
+
+            report = report.remove(to_remove, owner=self.args.owner)
+
+        return report
+
+
+def entry_point(cls=SetMaintenanceOff):
+    cls().main()
```
rohanpm: This is the statement missing test coverage.  Can you please fix that?  I guess you can do exactly as you did for SetMaintenanceOn case, which is covered.

JayZ12138: yeah, forget testing this.

pubtools/_pulp/tasks/set_maintenance/base.py
JayZ12138: yeah, this is better.

pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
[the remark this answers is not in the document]
JayZ12138: yeah, I prefer to go with option 2

tests/set_maintenance/test_set_maintenance.py
JayZ12138: I thought about it, but since I forget adding maintenance_report in the fake controller, if I want to test it here, I need to access the private attribute _maintenance_report of the fake client, which I shouldn't. 
And think more about it, what the set maintenance commands really do is calling kinds of APIs that pubtools.pulplib provides, it doesn't confirm if they work or not. It's pubtools.pulplib's job to make sure they work.

I can make a pull request to expose the maintenance report in fake controller and test here if you think it would make the tests more reliable.

rohanpm: > I thought about it, but since I forget adding maintenance_report in the fake controller, if I want to test it here, I need to access the private attribute _maintenance_report of the fake client, which I shouldn't.

Why do you need an extra method or any private attribute to get the maintenance report from the fake client/controller?  The public get_maintenance_report method ought to work, same as it would on a real client.

> And think more about it, what the set maintenance commands really do is calling kinds of APIs that pubtools.pulplib provides, it doesn't confirm if they work or not. It's pubtools.pulplib's job to make sure they work.

That's true, but it's the job of the code in pubtools-pulp to call those APIs correctly, and it's the job of the tests to verify that it indeed calls those APIs correctly, and the tests currently don't validate that at all.

I can see this by doing a little mutation testing.  I took this PR and made the following small patch:

```
diff --git a/pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py b/pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
index 42a839b..18154a4 100644
--- a/pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
+++ b/pubtools/_pulp/tasks/set_maintenance/set_maintenance_off.py
@@ -42,7 +42,7 @@ class SetMaintenanceOff(SetMaintenance):
             for repo_id in to_remove:
                 LOG.info(" - %s", repo_id)
 
-            report = report.remove(to_remove, owner=self.args.owner)
+            #report = report.remove(to_remove, owner=self.args.owner)
 
         return report
```

So, with this patch the maintenance-off command completely fails to remove any repos from maintenance mode.  However, all autotests still pass:

```
$ py.test tests/set_maintenance/test_set_maintenance.py -v
============================================= test session starts =============================================
platform linux -- Python 3.7.4, pytest-5.0.0, py-1.8.0, pluggy-0.12.0 -- /home/rmcgover/dev/python3/bin/python3
cachedir: .pytest_cache
rootdir: /home/rmcgover/src/pubtools-pulp, inifile: tox.ini
plugins: requests-mock-1.6.0, cov-2.7.1
collected 6 items                                                                                             

tests/set_maintenance/test_set_maintenance.py::test_maintenance_on PASSED                               [ 16%]
tests/set_maintenance/test_set_maintenance.py::test_maintenance_on_with_regex PASSED                    [ 33%]
tests/set_maintenance/test_set_maintenance.py::test_maintenance_on_with_repo_not_exists PASSED          [ 50%]
tests/set_maintenance/test_set_maintenance.py::test_maintenance_off PASSED                              [ 66%]
tests/set_maintenance/test_set_maintenance.py::test_maintenance_off_with_regex PASSED                   [ 83%]
tests/set_maintenance/test_set_maintenance.py::test_maintenance_off_with_repo_not_in_maintenance PASSED [100%]

========================================== 6 passed in 0.16 seconds ===========================================
```

This proves that the tests aren't effectively covering the behavior, since they wrongly pass even if the task implementation is completely broken.

The tests also fail to protect against an entire class of other bugs, such as: maintenance-off always unsets maintenance mode for *all* repos instead of the ones selected.

Getting the maintenance report from the client and verifying that the expected changes were made to it, will only cost a little extra code and make the tests much stronger.

PR #12, opened 2019-09-20
Refactor logging

rohanpm:
Includes two minor refactors to logging-related arguments:

- `--verbose` was dropped because it doesn't seem useful
- `--debug` was made more fine-grained so it can be used in more scenarios

See the included commits for more details.


pubtools/_pulp/task.py
```diff
@@ -89,14 +89,35 @@ def step(cls, name):
     def _basic_args(self):
         # minimum args required for a pulp CLI task
         self.parser.add_argument(
-            "--debug", action="store_true", help="show debug statements"
+            "--debug",
```
rajulkumar: "-d" argument could also be added here along with "--debug" so that multiple levels could be define with something like "-dd" or "-ddd" rather than writing "--debug" multiple times.

rohanpm: Yep, makes sense => https://github.com/release-engineering/pubtools-pulp/pull/14.

PR #13, opened 2019-09-20
Introduce publish command

rajulkumar:
This patch adds a new publish command that publishes
pulp repos to the defined endpints in the repo.

The command provides filters to fetch repos in pulp
or the repos provided as input given they are present
in pulp.

The repos can be published with publish options 'force'
to publish all the content and 'clear' to remove
content

The patch includes the unit tests and refernce doc


pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,221 @@
...
+    This command will publish the pulp repositories provided in the request or
+    fetched using the filters(url-regex or published-before) or an intersection
+    of input repos and filters.
+    """
+
+    def add_args(self):
+        super(Publish, self).add_args()
+
+        self.parser.add_argument(
+            "--repos", help="list of repos to be published", nargs="+"
```
rohanpm: maintenance-on/maintenance-off need the same arguments (select a set of existing repos), but they're named differently there:

```
                                    [--repo-url-regex REPO_URL_REGEX]
                                    [--repo-ids REPO_IDS [REPO_IDS ...]]
```

Could you please be consistent with it?  I don't think it makes sense to have one command taking ``--repo-ids`` and another taking ``--repos`` for the same purpose.

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,221 @@
...
+        )
+        self.parser.add_argument(
+            "--clean",
+            help="attempt to remove content not in the repo",
+            type=bool,
+            default=None,
```
rohanpm: I don't think clean and force work as you wanted now.  With this definition, you have to pass an argument for them:

```
  --clean CLEAN         attempt to remove content not in the repo
  --force FORCE         execute a full repo publish
```

Yet because ``bool(any_nonempty_string)`` is always True, it doesn't matter what value you pass.  For example ``--force false`` **enables** force.

Try putting ``action='store_true'`` instead, then the behavior is:

- default is False/None
- if you pass ``--clean`` (with no argument), it becomes ``True``

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,221 @@
...
+    def flush_ud(self, repos):
+        client = self.udcache_client
+        if not client:
+            LOG.info("UD cache flush is not enabled.")
+            return []
+
+        out = []
+        for repo in repos:
+            out.append(client.flush_repo(repo.id))
+            if repo.eng_product_id:
```
rohanpm: I don't think it's necessary to flush the products here - the old implementation doesn't do that.

(To be honest: I doubt it's necessary to flush the repos either.  Publishing the repos in Pulp shouldn't affect any of the data exposed via Pulp's API, so none of it needs flushing.  However, the old implementation does flush the repos, so it could be kept to guarantee compatible behavior until we know a specific reason to get rid of it.)

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,221 @@
...
+                out.append(client.flush_product(repo.eng_product_id))
+
+        return out
+
+    @step("Flush CDN cache")
+    def flush_cdn(self, repos):
```
rohanpm: This seems to be identical to clear_repo.  Think you could find a way to avoid duplicating the code?

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,221 @@
...
+            return list(repos.intersection(filtered_repos)) if repos else filtered_repos
+
+        return list(repos)
+
+    def _filtered_repo_distributors(self):
+        published_before = self.args.published_before
+        url_regex = self.args.url_regex
+
+        # define the criteria on available filters
+        if published_before and url_regex:
```
rohanpm: Arranging the code this way looks a little odd and not scalable - how will this code look if we add more distributor search clauses, like published_after?  Will there be one "if" case for every possible combination of arguments?

Would it rather be better to write the criteria for each branch just once, e.g.

```
crit = Criteria.true()
if published_before:
  crit = Criteria.and_([crit, Criteria.with_field("last_publish", Matcher.less_than(published_before)])
if url_regex:
  crit = Criteria.and_([crit, Criteria.with_field("relative_url", Matcher.regex(url_regex)])
```

Avoids having to write out the criteria for each argument more than once, and the number of ``if`` clauses is ``O(N)`` in terms of the arguments, instead of ``O(N^2)``.

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,221 @@
...
+                )
+                to_flush.append(flush_url)
+
+            LOG.debug("Flush: %s", to_flush)
+            flush = self.fastpurge_client.purge_by_url(to_flush)
+            return f_map(flush, lambda _: repo)
+
+        return [purge_repo(r) for r in repos if r.relative_url]
+
+    @step("Get repos")
```
rohanpm: It seems from the user's point of view, this step has the same meaning as "Check repos" in clear-repo command, so could it please be named the same?

If you think this step name is better, feel free to change the one in clear-repo.

tests/publish/test_publish.py
```diff
@@ -0,0 +1,314 @@
...
+            "--verbose",
+            "--repos",
+            "repo1",
+            "repo2",
+        ],
+    )
+
+
+def test_no_input_repos(command_tester):
+    """Fails if no repos are avialble to publish"""
```
rohanpm:
```suggestion
    """Fails if no repos are available to publish"""
```

tests/publish/test_publish.py
```diff
@@ -0,0 +1,314 @@
...
+    command_tester.test(
+        lambda: entry_point(FakePublish),
+        ["test-publish", "--pulp-url", "https://pulp.examle.com", "--verbose"],
```
rohanpm:
```suggestion
        ["test-publish", "--pulp-url", "https://pulp.example.com", "--verbose"],
```

tests/publish/test_publish.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+def test_nonexist_repos(command_tester):
+    """Fails when the requested repos doesn't exist"""
+    command_tester.test(
+        lambda: entry_point(FakePublish),
+        [
+            "test-publish",
+            "--pulp-url",
+            "https://pulp.examle.com",
```
rohanpm:
```suggestion
            "https://pulp.example.com",
```

tests/publish/test_publish.py
```diff
@@ -0,0 +1,314 @@
...
+    fake_publish = FakePublish()
+    fake_pulp = fake_publish.pulp_client_controller
+    _add_repo(fake_pulp)
+
+    command_tester.test(
+        fake_publish.main,
+        [
+            "test-publish",
+            "--pulp-url",
+            "https://pulp.examle.com",
```
rohanpm:
```suggestion
            "https://pulp.example.com",
```

tests/publish/test_publish.py
```diff
@@ -0,0 +1,314 @@
...
```
rohanpm:
```suggestion
            "https://pulp.example.com",
```

tests/publish/test_publish.py
```diff
@@ -0,0 +1,314 @@
...
```
rohanpm:
```suggestion
            "https://pulp.example.com",
```

tests/publish/test_publish.py
```diff
@@ -0,0 +1,314 @@
...
```
rohanpm:
```suggestion
            "https://pulp.example.com",
```

tests/publish/test_publish.py
```diff
@@ -0,0 +1,314 @@
...
```
rohanpm:
```suggestion
            "https://pulp.example.com",
```

tests/publish/test_publish.py
```diff
@@ -0,0 +1,314 @@
...
```
rohanpm:
```suggestion
            "https://pulp.example.com",
```

pubtools/_pulp/tasks/publish.py
rajulkumar: yeah. updated this.

pubtools/_pulp/tasks/publish.py
rajulkumar: +1

pubtools/_pulp/tasks/publish.py
rajulkumar: wasn't sure. so used it as in clear-repo. have changed it now to only flush repo

pubtools/_pulp/tasks/publish.py
rajulkumar: yeah. moved it to a different class.

pubtools/_pulp/tasks/publish.py
rajulkumar: agree. have made the changes as suggested.

pubtools/_pulp/tasks/publish.py
rajulkumar: I think "get repos" made more sense as this step actually gets the repos either for the repo ids provided or the repo ids fetched applying the filters. Same applies to clear-repo too, so updated it there.

JayZ12138: hm, but except get repos, you also checked if any repos' missing?

docs/publish.rst
```diff
@@ -0,0 +1,77 @@
...
+
+.. code-block::
+
+  pubtools-pulp-publish \
+    --pulp-url https://pulp.example.com/ \
+    --pulp-user admin \
+    --pulp-password XXXXX \
+    --repos my-repo1 my-repo2 ...
+
+Mentioned repositories will be pusblished to the defined
```
rohanpm:
```suggestion
Mentioned repositories will be published to the defined
```

pubtools/_pulp/task.py
```diff
@@ -143,3 +146,28 @@ def main(self):
 
         self.run()
         return 0
+
+
+class CDNCached(FastPurgeClientService):
+    # provding CDN cache related features
+
+    def _flush_cdn(self, repos):
```
rohanpm: - Should this perhaps be added to FastPurgeClientService class instead?  I guess in 99% of cases where we are getting a fastpurge client, we intend to use it to flush cache of some repos.  If you want to stick with introducing a separate class, could you please put that in a new file rather than stick it in the main task.py?
- I don't think it should be named with leading `_` - you certainly intend this method to be called from outside of this class.  Why give it a private name... what is the `_` supposed to communicate here?
- Should the `@step` decorator perhaps also be applied once here, rather than having to wrap the method in each place it's called?

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,190 @@
...
+LOG = logging.getLogger("pubtools.pulp")
+
+
+def publish_date(str_date):
+    # validates publish-before date
+    return datetime.strptime(str_date, "%Y-%m-%d")
+
+
+class Publish(PulpClientService, UdCacheClientService, PulpTask, CDNCached):
+    """Publishes the pulp repositories to the endpoints defined by the distributors
```
rohanpm: Let's try to keep the writing style consistent in the docs.

clear-repo https://release-engineering.github.io/pubtools-pulp/clear-repo.html for instance is:

> Remove all contents from one or more Pulp repositories.

Therefore shouldn't this be:

```suggestion
    """Publish one or more Pulp repositories to the endpoints defined by their distributors.
```

In particular the differences are:

- Written in the imperative mood ("Publish ..." and not "Publishes ...")
- Sentences end with a period `.`
- `pulp` is written as `Pulp` (capitalize proper nouns)

Ideally, the docs for different commands are not going to look different based on whichever dev happened to implement the command.

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,190 @@
...
+
+    This command will publish the pulp repositories provided in the request or
```
rohanpm:
```suggestion
    This command will publish the Pulp repositories provided in the request or
```

docs/publish.rst
```diff
@@ -0,0 +1,77 @@
...
+
+A typical invocation of publish would look like this:
+
+.. code-block::
+
+  pubtools-pulp-publish \
+    --pulp-url https://pulp.example.com/ \
+    --pulp-user admin \
+    --pulp-password XXXXX \
+    --repos my-repo1 my-repo2 ...
```
rohanpm: Docs don't match code.  The code is using `--repo-ids`.

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,190 @@
...
+        self.parser.add_argument(
+            "--repo-url-regex",
+            help="publish repos whose repo url match",
+            default=None,
+            type=re.compile,
+        )
+
+    def run(self):
+        to_await = []
+        LOG.info("run publish")
```
rohanpm: This doesn't look like a meaningful message for end-users.  Could it please be removed, changed to a DEBUG or changed to something useful?

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,190 @@
...
+        # flushing caches.
+        f_sequence(publish_fs).result()
+
+        # flush UD cache
+        to_await.extend(self.flush_ud(repos))
+
+        # flush CDN cache
+        to_await.extend(self.flush_cdn(repos))
+
+        # wait for everthing to finish.
```
rohanpm:
```suggestion
        # wait for everything to finish.
```

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,190 @@
...
+            LOG.info("Publishing %s", repo.id)
+            f = repo.publish(publish_opts)
+            f = f_map(f, partial(self.log_publish, repo))
+            out.append(f)
+
+        return out
+
+    def log_publish(self, repo, tasks):
+        # logs publish status
+        for task in tasks:
```
rohanpm: I don't think the logic here quite makes sense:

Most repos are going to have multiple distributors and so multiple tasks.  That means they're going to create multiple identical log messages in sequence, like:

      Publishing some-repo successful
      Publishing some-repo successful

Also I don't think you need to check task.succeeded because the API is already defined as only returning the tasks if they've succeeded (otherwise the future will raise an exception).

I'm thinking actually you don't need to log anything here at all and you can let the pubtools-pulplib task completion log messages cover it.  We discussed something similar in a maintenance-mode code reviews - currently those log messages may not contain all needed details, but it seems we can easily add them (task tags).

pubtools/_pulp/tasks/publish.py
rohanpm: I think that "check repos" is probably going to make more sense to the end-users.  Think about if you put a bank card in an ATM.  While that's loading data, is it more likely to say "Getting your account..." or "Checking your account..." ?  

It should be the latter, because the fact that some representation of the account/repo is being "gotten" from one place to another is an implementation detail which only makes sense if you think of the individual components of the system (i.e. that an ATM / Pub worker exists as one component which must "get" info from some separate component).  Ideally, users shouldn't be confronted with such details and it should appear more like one coherent system.

But, I mainly care about consistency, so I don't mind as long as the step is named the same in different tasks.

tests/logs/publish/test_publish/test_repo_publish_cache_cleanup.txt
```diff
@@ -0,0 +1,13 @@
+[    INFO] run publish
+[    INFO] Get repos: started
+[    INFO] Get repos: finished
+[    INFO] Publish: started
+[    INFO] Publishing repo1
+[    INFO] Publishing repo1 successful
```
rohanpm: This log shows up the complaint I raised earlier, about having identical log messages in sequence for a repo.  This doesn't make any sense for someone reading the log.

pubtools/_pulp/tasks/publish.py
rohanpm: Here's a PR extending the log messages in pulplib: https://github.com/release-engineering/pubtools-pulplib/pull/69

It would give you messages like:

     Pulp task completed: a1b2c3d4, repository:some-repo, action:publish

So I think if that PR is merged, then this logging could be dropped entirely.

pubtools/_pulp/task.py
rajulkumar: Idea here was to define a class of tasks that interact with CDN cache i.e. the tasks that make any modifications to CDN cache are task of type CDNCached. So added this class to task.py. The method with _ is available to flush CDN cache but be used within the definition/method specific to the task.
However, looking at the comments above, it might be better to have this as a generic class rather that being narrowed down to a task type. In that case, I'll put it in a new file with public method and step decorator.

pubtools/_pulp/tasks/publish.py
rajulkumar: +1

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,176 @@
...
+        url_regex = self.args.repo_url_regex
+
+        # define the criteria on available filters
+        crit = [Criteria.true()]
+        if published_before:
+            crit.append(
+                Criteria.with_field("last_publish", Matcher.less_than(published_before))
+            )
+        if url_regex:
+            crit.append(Criteria.with_field("relative_url", Matcher.regex(url_regex)))
```
rohanpm: I believe this is having exactly the same bug as fixed by https://github.com/release-engineering/pubtools-pulp/pull/15

Specifically, url_regex here is a compiled regex, and that can't be serialized to JSON, so it can't be used in a real search query.  Also, a pulplib PR to detect this bug was very recently landed https://github.com/release-engineering/pubtools-pulplib/pull/68, so I'd expect tests in this repo now to fail if run against latest pulplib.

Can you please look into it and fix it if it's indeed broken?

rohanpm: Yep, I retriggered the Travis build for this PR and this bug is indeed detected now by pubtools-pulplib >= 2.3.1, see https://travis-ci.org/release-engineering/pubtools-pulp/builds/592718772.

pubtools/_pulp/cdn_cache.py
```diff
@@ -0,0 +1,46 @@
...
+
+from pubtools._pulp.task import PulpTask
+from .services import FastPurgeClientService
+
+LOG = logging.getLogger("pubtools.pulp")
+
+step = PulpTask.step
+
+
+class CDNCache(FastPurgeClientService):
```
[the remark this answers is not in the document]
rajulkumar: I think that class is providing feature more in line with tasks (as one of the steps in the task or could be extended to a task) than service (parsing args and providing client). Also, moved CDNCache to common.py under tasks for this reason.

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # wait for the publish to complete before
+        # flushing caches.
+        f_sequence(publish_fs).result()
+
+        # flush UD cache
+        to_await.extend(self.flush_ud(repos))
+
+        # flush CDN cache
+        to_await.extend(self.flush_cdn(repos))
```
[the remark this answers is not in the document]
rajulkumar: Agree. That's how it's in current implementation too. However, it's otherwise in clear-repo(which is what was referred). Have changed the sequence as suggested.

pubtools/_pulp/tasks/publish.py
```diff
@@ -0,0 +1,176 @@
...
+            return []
+
+        out = []
+        for repo in repos:
+            out.append(client.flush_repo(repo.id))
+
+        return out
+
+    @step("Check repos")
+    def get_repos(self):
```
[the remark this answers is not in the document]
rajulkumar: done

pubtools/_pulp/tasks/publish.py
rajulkumar: Done

pubtools/_pulp/tasks/publish.py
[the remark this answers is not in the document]
rohanpm: CDN flush isn't really first with this update either because you never await completion of these CDN flushes before proceeding to UD.  I think that's fine though, note that we only flush cache for repomd.xml files which don't affect operation of UD at all (which only links directly to files under origin).

PR #15, opened 2019-09-30
fix couldn't search distributors while setting maintenance mode on

JayZ12138:
The reason why --repo-url-regex option isn't working is because,
for validation purpose, we add type=re.compile to that option, make
sure it's a valid pattern and passed it to make Criteria object directly,
which isn't supported.
Now passes RE.Pattern object's pattern to it.

### DIFF
--- a/pubtools/_pulp/tasks/set_maintenance/set_maintenance_on.py
+++ b/pubtools/_pulp/tasks/set_maintenance/set_maintenance_on.py
@@ -34,7 +34,7 @@ class SetMaintenanceOn(SetMaintenance):
         if self.args.repo_url_regex:
             # search distributors with relative_url, get the repo id from distributors
             crit = Criteria.with_field(
-                "relative_url", Matcher.regex(self.args.repo_url_regex)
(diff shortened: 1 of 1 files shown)


pubtools/_pulp/tasks/set_maintenance/set_maintenance_on.py
```diff
@@ -34,7 +34,7 @@ def adjust_maintenance_report(self, report):
         if self.args.repo_url_regex:
             # search distributors with relative_url, get the repo id from distributors
             crit = Criteria.with_field(
-                "relative_url", Matcher.regex(self.args.repo_url_regex)
+                "relative_url", Matcher.regex(self.args.repo_url_regex.pattern)
```
rohanpm: We can see here that no autotest update was needed to fix this, which is a concern as it means the same bug could happen again and nothing is protecting against it.

I think we should fix that in pulplib - https://github.com/release-engineering/pubtools-pulplib/pull/68/files.  That would avoid the need for an autotest here.

JayZ12138: Thanks for this :)

PR #17, opened 2019-10-08
Make repo-ids option accept comma separated argument

JayZ12138:
The advantage of accepting comma separated argument is, the command
fails immediately if space is used. In the opposite, if space is used
as the separator and if user wrongly used comma, then the command won't fail
until certain stage, or even succeeds with wrong results.

### DIFF
--- a/docs/publish.rst
+++ b/docs/publish.rst
@@ -18,7 +18,7 @@ A typical invocation of publish would look like this:
     --pulp-url https://pulp.example.com/ \
     --pulp-user admin \
     --pulp-password XXXXX \
-    --repo-ids my-repo1 my-repo2 ...
+    --repo-ids my-repo1,my-repo2 ...
 
 Mentioned repositories will be published to the defined
 endpoints in the distributors.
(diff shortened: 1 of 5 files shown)


rohanpm:
This makes sense to me but it doesn't match description of issue DELIVERY-7787, could you please update that to accurately reflect what's implemented?

JayZ12138:
> This makes sense to me but it doesn't match description of issue DELIVERY-7787, could you please update that to accurately reflect what's implemented?

Yeah. Updated JIRA description

JayZ12138:
> Please also put note in changelog as you changed cli API and it should be mentioned there.

I don't think we need to put it there, this repo is still under development and not really used anywhere. If there won't be a version change, why there should be changelog for the version?

rohanpm:
> > Please also put note in changelog as you changed cli API and it should be mentioned there.
> 
> I don't think we need to put it there, this repo is still under development and not really used anywhere. The Changelog is used for trace changes per versions, if there won't be a version change, why there should be changelog for the version?

I agree that changelog updates aren't strictly needed until version 1.0.0.  Maybe it'd be a good idea to change our processes to skip the changelog until that first stable version?  Right now it looks a little confusing, we have a changelog here but no stability is guaranteed until 1.0.0 anyway.

Actually, if this project was already used, this change couldn't be accepted since it's not backwards-compatible.

JayZ12138:
Guess I should merge this PR ASAP, so it won't block @rajulkumar releasing this repo.

PR #18, opened 2019-10-16
README: add badge for PyPI

rohanpm:
It's convenient to have a link directly to PyPI from the
project's README, to advertise that the project is available there
and also at a glance show the latest released version.


rohanpm:
Note most of our other projects used just a static link to PyPI - I'm giving the badge a try since it also shows the latest PyPI release, which is usually the main reason I'd be clicking through in the first place.

PR #30, opened 2021-02-09
Move to ubuntu:bionic env. of travis

rbikar:
Rust compiler isn't a part of the xenial
and is needed for building cryptography wheel
that is dependency of pyOpenSSL via edgegrid-python->fastpurge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: bionic
 sudo: false
 language: python
 before_install:


rbikar:
Example of failure https://travis-ci.com/github/rbikar/pubtools-pulp/jobs/481528148

rohanpm (approved):

PR #74, opened 2021-12-16
push: introduce --pre-push option

rohanpm:
This option enables pre-push mode (sometimes also called 'nochannel').
In this mode the workflow performs only those steps which load content
into Pulp without making it visible to end-users.

### DIFF
--- a/pubtools/_pulp/tasks/push/command.py
+++ b/pubtools/_pulp/tasks/push/command.py
@@ -76,7 +76,15 @@ class Push(
 
         self.add_publisher_args(self.parser)
 
-        # TODO: --skip
+        self.parser.add_argument(
+            "--pre-push",
+            action="store_true",
+            help=(
+                "Pre-push mode: do as much as possible without making content "
+                "available to end-users, then stop. May be used to improve the "
+                "performance of a subsequent full push."
+            ),
+        )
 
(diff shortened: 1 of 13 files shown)


pubtools/_pulp/tasks/push/command.py
```diff
@@ -174,14 +186,24 @@ def uploaded_items(self, items):
...
+            "%s uploading" % len(needs_upload),
+        ]
 
-        # Anything already in the system can be immediately yielded.
-        for item in uploaded:
+        if self.args.pre_push:
+            messages.append("%s skipped during pre-push" % len(prepush_skipped))
+            event["items-prepush-skipped"] = len(prepush_skipped)
+
+        LOG.info("Upload items: %s", ", ".join(messages), extra={"event": event})
```
negillett, line 203: This introduces some inconsistency in log message format. Are we to assume Update will get the same treatment?

negillett (approved):

rajulkumar (approved):

pubtools/_pulp/tasks/push/command.py
rohanpm: Yes, it'd be a good idea to make them consistent in a later patch.

PR #76, opened 2021-12-21
Introduce fix-cves command

rajulkumar:
Implements the fix-cves workflow which:
- updates the given cves to the advisory
- uploads the updated advisory to the nochannel repo
- publish all the repos in advisory's repo membership
- flush Akamai and UD cache, for the provided urls


tests/logs/fix_cves/test_fix_cves/test_noop_on_same_cves.txt
```diff
@@ -0,0 +1,7 @@
+[    INFO] Get erratum: started
+[    INFO] Get erratum: finished
+[    INFO] Compare CVEs: started
+[    INFO] CVEs from erratum: ['CVE-123']
+[    INFO] CVEs for fix: ['CVE-123']
+[    INFO] New CVEs are same as the ones on the advisory. Nothing to do.
```
rohanpm, line 6: As mentioned elsewhere - I don't think this should really be a no-op.  It should still go ahead and publish the repos.

rohanpm (changes requested):

pubtools/_pulp/tasks/fix_cves.py
```diff
@@ -0,0 +1,171 @@
...
+
+        # compare new cves with erratum cves
+        if self.same_cves(erratum, cves):
+            return
+
+        # get repo memberships from erratum
+        repos = self.get_affected_repos(erratum)
+
+        # process erratum
+        self.process_erratum_for_upload(erratum, cves)
```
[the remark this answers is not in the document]
rajulkumar: Thanks. Fixed this and added the tests to verify.

pubtools/_pulp/tasks/fix_cves.py
```diff
@@ -0,0 +1,171 @@
...
+        return [r for r in resp.result()]
+
+    def get_nochannel_repo_id(self, repo_type):
+        return NOCHANNEL_REPO_MAP[repo_type]
+
+    @step("Get affected repos")
+    def get_affected_repos(self, erratum):
+        # get repos listed in repository memberships
+        repo_ids = erratum.repository_memberships
+        repo_ids = set(repo_ids) - set(NOCHANNEL_REPO_MAP.values())
```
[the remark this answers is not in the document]
rajulkumar: Removing this return the nochannel repos too in the list of affected repos. That list is later used to pick one of the random repos to upload which may return the nochannel repo as the destination to upload. However, it will definitely reduce the serialization.

pubtools/_pulp/tasks/fix_cves.py
```diff
@@ -0,0 +1,167 @@
...
+
+        cves = [cve_id.upper() for cve_id in cves]
+        LOG.info("CVEs for fix: %s", ", ".join(sorted(cves)))
+
+        # compare with new cves
+        if set(erratum_cves) == set(cves):
+            LOG.info(
+                "New CVEs are same as the ones on the advisory. Nothing to update."
+            )
+            return 1
```
rohanpm, line 114: Minor - How about returning booleans rather than using 1/0 for true/false?

pubtools/_pulp/tasks/push/items/erratum_conv.py
```diff
@@ -46,27 +46,34 @@ def unit_for_item(item, old_unit):
...
-
     # The version from the push item is only used if there was no erratum in pulp,
     # or the pulp version couldn't be bumped.
+    if old_unit:
+        return bump_erratum_version(old_unit, item.version)
+
     return item.version
 
 
+def bump_erratum_version(unit, default_version=1):
```
rohanpm, line 57: Currently the way you've written this, it will return a str in the successful bump case and an int in the "failed to bump" case.

Would it be better to return a consistent type?

docs/fix-cves.rst
```diff
@@ -0,0 +1,50 @@
...
+
+  pubtools-pulp-fix-cves \
+    --pulp-url https://pulp.example.com/ \
+    --pulp-user admin \
+    --pulp-password XXXXX \
+    --advisory RHXA-123:45 \
+    --cves CVE-123,CVE-345
+
+Mentioned CVEs will be updated in the advisory and uploaded
+to the nochannel repo. The repos this advisory is a part will
```
rohanpm, line 25: Could you please drop the references to nochannel repo in the docs?

pubtools/_pulp/tasks/fix_cves.py
```diff
@@ -0,0 +1,167 @@
...
+        LOG.info("Uploaded to %s", repo.id)
+
+    def get_repo_from_pulp(self, repo_id):
+        crit = Criteria.with_id(repo_id)
+        resp = self.pulp_client.search_repository(criteria=crit)
+
+        return [r for r in resp.result()]
+
+    def get_upload_repo(self, repos):
+        return random.choice(repos)
```
rohanpm, line 88: I see that using random might pose a stability issue for tests and maybe this is why you mocked it.

May I suggest the following instead: rather than random.choice(), create a random.Random object with a controllable seed, e.g. like this:

```python
self.random = random.Random(os.getenv('PUBTOOLS_SEED') or random.random())
```

Then call self.random.choice.

Tests can then obtain results which are stable but still random by setting PUBTOOLS_SEED to some specific value before constructing an instance of the task, e.g. just '1'.  This also works fine from higher level tests (e.g. if there are some Pub tests which run these tasks).

tests/fix_cves/test_fix_cves.py
```diff
@@ -0,0 +1,259 @@
...
+
+    controller.insert_repository(repo)
+    controller.insert_repository(nochannel_repo)
+    controller.insert_units(repo, [erratum])
+    controller.insert_units(nochannel_repo, [erratum])
+
+
+@pytest.fixture(autouse=True)
+def fake_random():
+    with patch("pubtools._pulp.tasks.fix_cves.random.choice") as mock_choice:
```
rohanpm, line 133: Elsewhere I put a suggestion which would allow you to avoid mocking this.

rohanpm (changes requested):

pubtools/_pulp/tasks/fix_cves.py
rajulkumar: Done

pubtools/_pulp/tasks/push/items/erratum_conv.py
rajulkumar: Agree. Done

docs/fix-cves.rst
rajulkumar: Done

rohanpm (approved):

rohanpm (approved):

PR #80, opened 2022-01-18
Add pulp flush  hook [RHELDST-7309]

negillett:
A hook, "task_pulp_flush", is now specified and called when
pubtools-pulp-push completes publishing all repositories. Hook
implementations may now subscribe to these events to flush pulp-derived
caches and datastores.

### DIFF
--- a/pubtools/_pulp/hooks.py
+++ b/pubtools/_pulp/hooks.py
@@ -1,0 +1,17 @@
+import sys
+
+from pubtools.pluggy import hookspec, pm
+
+
+@hookspec
+def task_pulp_flush():
+    """Invoked during task execution after successful completion of all Pulp
+    publishes.
+
+    This hook is invoked a maximum of once per task, to indicate that all Pulp
+    content associated with the task is considered fully up-to-date. The
+    intended usage is to flush Pulp-derived caches or to notify systems that
+    Pulp content may have recently changed."""
+
+
+pm.add_hookspecs(sys.modules[__name__])
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
             "pubtools-pulp-maintenance-off = pubtools._pulp.tasks.set_maintenance.set_maintenance_off:entry_point",
             "pubtools-pulp-publish = pubtools._pulp.tasks.publish:entry_point",
             "pubtools-pulp-push = pubtools._pulp.tasks.push:entry_point",
-        ]
+        ],
     },
     project_urls={
         "Documentation": "https://release-engineering.github.io/pubtools-pulp/",
--- a/pubtools/_pulp/tasks/common.py
+++ b/pubtools/_pulp/tasks/common.py
@@ -10,6 +10,8 @@ from pubtools.pulplib import PublishOptions
 from pubtools._pulp.task import PulpTask
 from pubtools._pulp.services import FastPurgeClientService, UdCacheClientService
 
+from ..hooks import pm
+
 LOG = logging.getLogger("pubtools.pulp")
 
 step = PulpTask.step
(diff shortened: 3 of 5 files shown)


pubtools/_pulp/hooks.py
```diff
@@ -0,0 +1,12 @@
+import sys
+
+from pubtools.pluggy import hookspec, pm
+
+
+@hookspec
+def task_pulp_flush():
+    """Invoked during task execution after successful completion of all
```
rohanpm, line 8: Would you mind extending the doc string a little?  A concrete suggestion is below.

The main points are:
- the "all" Pulp writes unfortunately turns out to be false - looking at the code I can see now that we have to set cdn_published field after this hook is invoked.  Annoying, but the code is right so the doc string should be adjusted.
- probably good to say what's the intended usage of it (similarly as mentioned in the commit message)

So my attempt at a doc string would be:

```
Invoked during task execution after successful completion of all Pulp publishes.

This hook is invoked a maximum of once per task, to indicate that all Pulp content
associated with the task is considered fully up-to-date. The intended usage is to
flush Pulp-derived caches or to notify systems that Pulp content may have recently
changed.
```

By the way, this ends up at https://release-engineering.github.io/pubtools/hooks.html#hook-reference in case it wasn't clear.

negillett:
It seems we'll need https://github.com/release-engineering/pubtools-pulplib/pull/156 merged then https://github.com/release-engineering/pubtools-exodus/pull/2 and then an pubtools-exodus release to PyPI.

@rohanpm, granted that's correct, should I create a Jira issue to add PyPI deployment to pubtools-exodus' tox?

pubtools/_pulp/hooks.py
```diff
@@ -0,0 +1,17 @@
...
+
+@hookspec
+def task_pulp_flush():
+    """Invoked during task execution after successful completion of all Pulp
+    publishes.
+
+    This hook is invoked a maximum of once per task, to indicate that all Pulp
+    content associated with the task is considered fully up-to-date. The
+    intended usage is to flush Pulp-derived caches or to notify systems that
+    Pulp content may have recently changed.."""
```
rohanpm, line 14:
```suggestion
    Pulp content may have recently changed."""
```

setup.py
```diff
@@ -48,7 +48,10 @@ def get_requirements():
             "pubtools-pulp-maintenance-off = pubtools._pulp.tasks.set_maintenance.set_maintenance_off:entry_point",
             "pubtools-pulp-publish = pubtools._pulp.tasks.publish:entry_point",
             "pubtools-pulp-push = pubtools._pulp.tasks.push:entry_point",
-        ]
+        ],
+        "pubtools.hooks": [
```
rohanpm, line 52: This part is not right, this should live in pubtools-exodus.

rohanpm:
> It seems we'll need [release-engineering/pubtools-pulplib#156](https://github.com/release-engineering/pubtools-pulplib/pull/156) merged then [release-engineering/pubtools-exodus#2](https://github.com/release-engineering/pubtools-exodus/pull/2) and then a pubtools-exodus release to PyPI.
> 
> @rohanpm, granted that's correct, should I create a Jira issue to add PyPI deployment to pubtools-exodus' tox?

Those parts need to happen eventually but should not be connected to this PR.  There should be nothing blocking this PR going first.

If you meant because CI now fails after adding the entry point stuff, I believe that's because it was added to the wrong project.  pubtools-pulp should not have to "know" that pubtools-exodus implements some hooks.  The entry point should be added on pubtools-exodus.  Once removed from here you should see the CI problem go away.

rohanpm (changes requested):

setup.py
negillett: ah, okay, thanks.

rohanpm (approved):

rajulkumar (approved):

PR #90, opened 2022-02-20
Refactor push workflow into loosely coupled phases [RHELDST-8331]

rohanpm:
Each step in the push workflow was previously implemented via
generators. That approach limits concurrency as much of the push
logic still takes place in the main thread which can do only one
thing at a time. For example, it is ideally possible for the "calculate
checksums" step to continue calculating checksums of files while the
"query pulp" step is already issuing Pulp queries, but that couldn't be
done with the generator approach.

This commit fixes the problem by refactoring each step of the workflow
into a "phase", which runs on its own thread and communicates with
previous/later phases via queues. It brings several advantages:

- removes unnecessary sources of blocking, as each phase runs at its own
  pace without directly blocking the others

- smarter batching of items becomes possible

- new code is arguably more readable since each phase has clearly
  delineated inputs and outputs

- as queue sizes can be tracked, the progress across phases can be
  monitored and visualized - useful for long-running pushes. (Not
  included in this commit but coming soon)

One of the downsides is that a specific start/end order of each executed
step can no longer be guaranteed. For these reason, some log-based tests
had to be dropped.


pubtools/_pulp/tasks/push/items/base.py
```diff
@@ -140,6 +145,146 @@ def upload_context(cls, pulp_client):
         """
         return UploadContext(client=pulp_client)
 
+    @classmethod
```
rohanpm, line 148: Note, all this code has just been moved from command.py but otherwise has no substantial changes.

rbikar (approved):
@rohanpm  lgtm, btw any plans on adding unit tests for Context, Phase and its subclasses? I know that the code is covered by near-to-integration tests, I guess smaller unit tests might be  useful while fixing some subtle bugs or while adding new functionality to some phase.

rohanpm:
@rbikar my strategy is generally to get as much coverage as possible with tests which operate at a similar level as end-users.  Then drop down to unit tests to get the coverage for edge cases which were difficult to reach with the higher-level tests.

So far it seems like that wasn't needed here, as all the code changes already hit 100% coverage just with the existing tests.

I mentioned something in the commit message about adding some progress monitoring of the queues in a later commit.  I can make it a point to add some unit tests along with that, since it doesn't seem like the higher-level tests would be able to verify that beyond "it doesn't crash".

rbikar:
> @rbikar my strategy is generally to get as much coverage as possible with tests which operate at a similar level as end-users. Then drop down to unit tests to get the coverage for edge cases which were difficult to reach with the higher-level tests.
> 
> So far it seems like that wasn't needed here, as all the code changes already hit 100% coverage just with the existing tests.
> 
> I mentioned something in the commit message about adding some progress monitoring of the queues in a later commit. I can make it a point to add some unit tests along with that, since it doesn't seem like the higher-level tests would be able to verify that beyond "it doesn't crash".

Fair enough.

PR #100, opened 2022-03-02
push: add checks on the outcome of item uploads & updates

rohanpm:
When we upload or update an item in Pulp, even if the Pulp API says it
succeeded, we should still double-check the new state of the item to
verify that it had the desired impact. (i.e. if we uploaded an item then
it should be in at least one repo; if we updated an item then it should
not require any further updates.

Fixes a couple of TODOs.

### DIFF
--- a/pubtools/_pulp/tasks/push/items/base.py
+++ b/pubtools/_pulp/tasks/push/items/base.py
@@ -410,8 +410,21 @@ class PulpPushItem(object):
 
         upload_tasks = f_flat_map(repo_f, self.upload_to_repo)
 
-        # TODO: should we check here and raise if refresh didn't find that we're present?
-        return f_flat_map(upload_tasks, lambda _: self.with_pulp_refreshed(ctx.client))
+        # Helper to verify that we've really got into at least one repo as a result
+        # of the upload.
+        def asserting_uploaded_ok(item):
+            if not item.in_pulp_repos:
+                msg = (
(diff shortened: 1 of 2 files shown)


pubtools/_pulp/tasks/push/items/base.py
```diff
@@ -410,8 +410,21 @@ def ensure_uploaded(self, ctx, repo_f=None):
 
         upload_tasks = f_flat_map(repo_f, self.upload_to_repo)
 
-        # TODO: should we check here and raise if refresh didn't find that we're present?
-        return f_flat_map(upload_tasks, lambda _: self.with_pulp_refreshed(ctx.client))
+        # Helper to verify that we've really got into at least one repo as a result
+        # of the upload.
+        def asserting_uploaded_ok(item):
```
negillett, line 415: minor: should these methods just be "assert_uploaded/updated", or are those name taken?

negillett (approved):

rajulkumar (approved):

pubtools/_pulp/tasks/push/items/base.py
rohanpm: It would be unusual for an 'assert' function or statement to return a value.  In this case we need a function which does a check on a value, but then also returns it.  The "ing" suffix is meant to indicate this, like: "given some value, give me the same value back, while asserting that some condition is satisfied".

PR #115, opened 2022-03-14
Introduce delete command

rajulkumar:
Implements the delete command
- which is common for different unit types
- supports deletion of rpms, srpms, modules, isos
- treats all the input files as "files" irrespective of type
- validates the filename for rpm/srpm and module type
- supports deletion of packages and modules from advisories
- depends on the advisory info from Pulp rather than Errata Tool


pubtools/_pulp/tasks/common.py
```diff
@@ -158,3 +169,87 @@ def publish_with_cache_flush(self, repos, units=None):
         out.append(flush_ud)
 
         return out
+
+
+class PushItemRecorder(CollectorService):
```
rohanpm, line 174: The functionality here is specifically about deleting, and the names of methods can be misleading if neither the class nor method names indicate this. e.g. nobody would guess that PushItemRecorder.push_item_for* methods always return a DELETED push item and aren't appropriate to call in other cases.  Do you think the class could be given a name which is more specifically about deletion?

pubtools/_pulp/tasks/delete/base.py
```diff
@@ -0,0 +1,124 @@
+from ast import Del
```
rohanpm, line 1: This must be a mistake, ast module is for parsing Python.

pubtools/_pulp/tasks/delete/__init__.py
```diff
@@ -0,0 +1,15 @@
+from .base import multi_dispatch
+from .delete_packages import DeletePackages
+from .delete_advisory import DeleteAdvisory
+
+
+Delete = multi_dispatch("Delete", DeleteAdvisory, DeletePackages)
```
rohanpm, line 6: I'm not sure if the multi_dispatch thing is the way to go here.  Could we just do all deletions that the user asked for?  i.e. why can't I delete an advisory and also delete some loose files at once?  It seems a bit of an artificial restriction that the code has to choose either to jump into a class which only supports deleting an advisory, or only supports deleting standalone files.

pubtools/_pulp/tasks/common.py
rajulkumar: Agree. Need to modify this class for the reason you mentioned and extend this for handling of pushitems from sources other than tasks.

pubtools/_pulp/tasks/delete/base.py
rajulkumar: Yes, removed.

pubtools/_pulp/tasks/delete/__init__.py
rajulkumar: I thought the same and then went ahead to split into two different classes as there's a difference in the approach of the two processes i.e. deleting standalone units and packages from advisory. However, the dispatch part is not working as expected as it looks like there's a deadlock as making the the dispatch decision and to instantiation of the class should be done on paring the args which again requires some instance of the class to utilize the existing parser in PulpTask and other parent classes.
So I believe going by the suggested approach to handle all the types in the same class makes sense. This will also give an option to do a mix-n-match deletion as mentioned above.
Taking this approach, can you please suggest if it should be an all-or-none removal or whatever applies for the repos to be removed from i.e. If a unit belong to two out of three provided repos, should the removal be skipped as the unit doesn't belong to all the repos or the unit be removed from the repos where it's present? Existing implementation in Pub has a mix of both.

rohanpm: > If a unit belong to two out of three provided repos, should the removal be skipped as the unit doesn't belong to all the repos or the unit be removed from the repos where it's present?

I suggest the latter - it should be removed from the repos where it's present.  Otherwise the command wouldn't be retryable if interrupted partway through (e.g. imagine you ask it to remove a file from 3 repos, it succeeds from 2 and then crashes - you should be able to just retry your command without having to adjust the repo list to include just the final repo).

I think the command should be thought of more as "ensure content is absent", rather than "delete content".  So it's not an error if the content is already absent from some of the specified repos.  Like `rm -f somefile` does not fail if the file already is missing.

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,709 @@
...
+        advisory = self.args.advisory
+
+        if not (self.args.file or self.args.advisory):
+            self.fail("One of --file or --advisory is required")
+
+        # separates files into rpms, isos and modules
+        rpms, files, modules, _ = self.separate_files(self.args.file or [])
+
+        # collect skipped items
+        # TODO
```
rohanpm, line 124: What does this mean actually?

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,709 @@
...
+            module_f = self.delete_modules(modules, signing_keys, repo_names)
+
+        # delete packages from advisory
+        if advisory:
+            adv_f = self.delete_from_advisory(advisory, repo_names)
+
+        # wait for the futures to resolve
+        # each future returns a list of futures for content removal on each repo
+        fs = f_zip(rpm_f, file_f, module_f, adv_f)
+        fs = [f for sublist in fs.result() if sublist for f in sublist]
```
rohanpm, line 145: I find this tricky to read.  Would it be simpler if you initialized the futures like rpm_f as `f_return([])` instead of just `f_return()` ?  If methods are designed to return a list, you may as well use empty lists to represent "nothing was removed" rather than None.

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,709 @@
...
+            for f in self.to_await:
+                f.result()
+
+    def delete_from_advisory(self, advisory, repos):
+        pushitems = []
+        advisory_id = None
+
+        # get errata from source
+        LOG.info("Get items from source")
+        with Source.get(advisory) as source:
```
rohanpm, line 167: Not sure about this.  This is going to query ET and get push items on the advisory, as it exists today.

But I'd guess one of the main use-cases of this command is to deal with the side-effects of an advisory *changing*.  Let's say an advisory originally has packages A, B, C, D and a push is done.  Then the advisory is modified so it only contains packages A, B.  That's not going to "unpush" packages C, D, so I think that scenario is a case where someone may want to use this command in order to remove whatever was pushed by an earlier version of the advisory.  But that's not going to work if the code is based around looking at the *current* version of the advisory to decide what should be done.

Does it work this way on the old code? Deleting only what's in the advisory as fetched from ET during the task?

I think the command should work without querying ET and should just remove whatever's in Pulp associated with the advisory.

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,709 @@
...
+                continue
+            mapped = False
+            for repo in repos or item.dest:
+                if repo in item.dest:
+                    repo_map.setdefault(repo, []).append(item)
+                    mapped = True
+
+            if not mapped:
+                LOG.warning(
+                    "%s is not present in any repo: %s", item.name, ", ".join(repos)
```
rohanpm, line 308: This warning isn't right: item.dest isn't telling you which repos the item is *present in*.  It's telling you which repos Errata Tool (or other item source) wants the item to be pushed to.  I think it's a symptom of the overall concept here being wrong, the command should be driven only by the arguments and the state of Pulp, not the state of ET.

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,709 @@
...
+            RpmUnit, self._rpm_search_criteria(rpm_names, signing_keys)
+        )
+        rpms_f = self.search_content(criteria)
+        rpms_f = f_map(
+            rpms_f,
+            partial(self.log_missing_units, unit_names=rpm_names, unit_attr="filename"),
+        )
+        return rpms_f
+
+    @step("Get ISOs")
```
rohanpm, line 364: Could you please call this "files" instead of "ISOs" ?  We should try to move away from the "ISO == generic file" terminology which doesn't make sense.

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,709 @@
...
+        if missing:
+            missing = sorted(list(missing))
+            LOG.warning("Requested unit(s) doesn't exist: %s", ", ".join(missing))
+
+        LOG.info("%s unit(s) found for deletion", len(found))
+        # TODO: collect missing units "NOT FOUND"
+
+        return searched_units
+
+    @step("Mapping units to repo")
```
rohanpm, line 464: I don't think this should be a step, there's nothing asynchronous here at all, won't it always complete in less than a second?

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,709 @@
...
+            missing = ", ".join(sorted(list(missing)))
+            LOG.warning("No units to remove from %s", missing)
+
+        return repo_map, unit_map
+
+    @step("Remove RPMs")
+    def remove_rpms(self, repo_map):
+        return self.delete_content(RpmUnit, repo_map, self._rpm_remove_crit)
+
+    @step("Remove ISOs")
```
rohanpm, line 495: Can we say files instead of ISOs?

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,709 @@
...
+                    Criteria.with_field("filename", item.name),
+                    Criteria.with_field("signing_key", item.signing_key.lower()),
+                )
+            )
+
+        criteria = Criteria.and_(Criteria.with_unit_type(RpmUnit), Criteria.or_(*crit))
+
+        return repo.remove_content(criteria=criteria)
+
+    @step("Sort out files")
```
rohanpm, line 572: I don't think this should be a step either, it should complete more or less instantaneously.

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,709 @@
...
+    def separate_files(self, filenames):
+        rpms = []
+        modules = []
+        files = []
+        skipped = []
+        for f in filenames:
+            if f.endswith(".rpm"):
+                rpms.append(f)
+            elif f.endswith(".iso"):
```
rohanpm, line 581: This looks wrong.  Nothing in here should be limited to removal of .iso files specifically.  "ISO" repositories in Pulp contain files not matching this pattern, so this would prevent the command from handling them.

I think the approach is just flawed actually.  You can't possibly know ahead of time if a certain file is going to be present in Pulp as an rpm unit, iso unit or something else just based on the filename.  For example: it's possible (maybe by accident) for someone to take an RPM file and upload it as a generic file (iso) into a FileRepository.  If that happened, then this command is the one we'd want to use to fix the situation and delete the file from the repo so it can be repushed to the correct repo.  But logic like "if filename ends in .rpm then we'll only look for it in RPM repositories" would prevent that.

If the user passed IDs of any FileRepository, I think we need to be searching for *all* listed filenames in that repository, even if 99%+ of the time we won't find RPMs there.

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,709 @@
...
+    """Represents a unit with destination repos"""
+
+    unit = attr.ib()
+    """A Pulp unit"""
+
+    repos = attr.ib()
+    """The list of repos from where the unit to be removed"""
+
+
+class Delete(PulpClientService, CollectorService, Publisher, PulpTask):
```
rohanpm, line 63: Some of the methods in here look very similar or maybe copy-pasted from ClearRepo class.  Would it be possible to cut down on the code duplication?

rohanpm (changes requested):

pubtools/_pulp/tasks/delete.py
rajulkumar: Makes sense. Pulp should be the source-of-truth. However, current implementation gets the items from advisory source, verify that input repos are in item.dest(remove others in item.dest), map items to the repos in the input and remove those elements from the mapped repos.
IIUC, it should get the advisory from Pulp, gather the packages from the advisory and then remove them from the requested repos? Currently, only packages from the advisory is removed. Should it be extended to remove all the items i.e. both packages and modules in the advisory pkglist?

pubtools/_pulp/tasks/delete.py
rajulkumar: I see what you mean. So it should look for all the requested files in the requested repos(file and yum repos only) and remove whatever is found in each repo (generating the criteria for removal based on the Unit type)?
I think this should be better with the available apis like repository.search_content rather than searching all the files in Pulp, verifying the requested repos against their repository_memberships and then removing the matched content from the repos.

pubtools/_pulp/tasks/delete.py
rajulkumar: Yes, will move the code to a common place to be used by both the commands.

pubtools/_pulp/tasks/delete.py
rajulkumar: Ack.

pubtools/_pulp/tasks/delete.py
rajulkumar: Sure

pubtools/_pulp/tasks/delete.py
rajulkumar: Ack

pubtools/_pulp/tasks/delete.py
rajulkumar: Sure.

pubtools/_pulp/tasks/delete.py
rajulkumar: Yes, might not need this if taking the above suggested approach.

pubtools/_pulp/tasks/delete.py
rajulkumar: Ack

pubtools/_pulp/tasks/delete.py
rajulkumar: There are some files that are skipped or not found during the process. Should we send those files to collector or logging them is enough?
This is a TODO to collect those skipped files.

rohanpm: That's rather open-ended I'd say.  I don't think much thought was put into the semantics of "skipped" in pub originally and I don't think we have to stick with it.  I'd say that it's OK to err on the side of less work / less complexity, and just leave it out.

pubtools/_pulp/tasks/delete.py
rohanpm: If the current implementation is initializing the list of files by querying ET, then I'm unsure if this needs to change.  Though I believe that behavior doesn't make sense, it's not meant to be in scope here to fix all existing problems with the command.  I'd like to review some past uses of delete-packages-from-adv and see why it was used, and judge whether the command really did what the users wanted it to.  So let me have a bit more time to reply to this.

About whether it should handle modules too.  If it were a new command, I'd say yes and it doesn't make sense that it omits modules.  But again it wasn't the intended scope of this piece of work to fix all existing problems with the command.  What's your opinion though?  It also depends on how difficult it would be.

pubtools/_pulp/tasks/delete.py
rohanpm: Yes, for example, it could be done by:

- if user requested delete of "some-file", then:
   - always do a search for FileUnit with path="some-file"
   - if "some-file" can be parsed as an RPM filename, then also do a search for RpmUnit using the parsed value
   - if "some-file" can be parsed as an NSVCA, then also do a search for ModulemdUnit using the parsed value
- when you've got all matching units, you can look at their repository_memberships, look at the repos passed in by the user and plan out what needs to be deleted.

pubtools/_pulp/tasks/delete.py
rajulkumar: I think it shouldn't be much of a work. We can get the list of module nsvca from the erratum pkglist and pass it to the delete_modules method to take care of the removal.

rohanpm: About whether it should figure out the files for deletion by querying ET or by only querying Pulp.

I tried looking into historical tasks, but there's not a lot of data either way since the command only seems to be used a couple of times per year on average.  Still, I did find one example which shows how the concept of querying ET is flawed:

- here's a delete task: https://url.corp.redhat.com/e050479
- an example what it queried from ET: delete aap-azure-ui-0.0.1-0.1.20220221.git0a32a2d.el8ap.noarch.rpm from ansible-automation-platform-2_DOT_1-for-rhel-8-x86_64-rpms
- but ET content today: https://url.corp.redhat.com/14a02fb
- shows different repo(s) for the RPMs

In that case it worked out OK, but it seems error-prone: if someone had updated the repos in ET before the delete command was run, then the delete command wouldn't have found the correct repos for deletion.  I doubt that anyone is aware of that, and even if it hasn't bitten the users yet, it surely will eventually.

Combined with my opinion that, if the code were changed to query Pulp instead of ET then it would likely be not significantly more complex or maybe even simpler... yeah, I think it'd be best to revise it so it's just driven by Pulp state and doesn't call out to ET at all.

rajulkumar: Thanks. I updated the delete advisory workflow to fetch the details from Pulp, gather the packages and modules name and then pass them to delete_rpms and delete_modules for deletion. Also, it supports deleting multiple advisories now instead of one as in the current implementation.

pubtools/_pulp/tasks/delete.py
rajulkumar: Removed it as it wasn't required with the new workflow.

pubtools/_pulp/tasks/delete.py
rajulkumar: Done

pubtools/_pulp/tasks/delete.py
rajulkumar: Removed it as there's nothing asynchronous.

pubtools/_pulp/tasks/delete.py
rajulkumar: Done

pubtools/_pulp/tasks/delete.py
rajulkumar: Removed it

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,735 @@
...
+        self.parser.add_argument(
+            "--signing-key",
+            help="remove the content with these signing-keys(s)",
+            type=str,
+            action=SplitAndExtend,
+            split_on=",",
+        )
+
+        self.parser.add_argument(
+            "--no-publish",
```
rohanpm, line 108: Can this please be implemented as `--skip publish` instead?  That's how it works in clear-repo command - https://release-engineering.github.io/pubtools-pulp/clear-repo.html#example-skipping-publish

In fact if you use a `@step("Publish")` somewhere, you don't have to implement any code explicitly for the skipping, because `@step` does it for you.

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,735 @@
...
+        # each advisory
+        file_fs.append(file_f)
+        fs = f_sequence(file_fs)
+        cleared_repos = [cr for sublist in fs.result() for cr in sublist]
+        for cr in cleared_repos:
+            repo = cr.repo
+            repos.setdefault(repo.id, repo)
+
+        # publish affected repos
+        if repos and not self.args.no_publish:
```
rohanpm, line 154: This should probably be divided into steps.

- This part should go into a `@step("Publish")`, that would also help you get `--skip publish` working.
- Most of the steps above should probably go into a `@step("Delete content")` or similar.

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,735 @@
...
+        for unit, runit_item in sorted(unit_map.items()):
+            repos = runit_item.repos
+            if not repos:
+                missing.append(unit)
+            else:
+                LOG.info("Deleting %s from %s", unit, repos)
+
+        if missing:
+            LOG.warning(
+                "Unit(s) doesn't belong to any requested repos %s: %s",
```
rohanpm, line 290:
```suggestion
                "Unit(s) don't belong to any requested repos %s: %s",
```

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,735 @@
...
+        found = []
+
+        for unit in searched_units:
+            found.append(getattr(unit, unit_attr))
+
+        missing = set(unit_names) - set(found)
+        if missing:
+            missing = sorted(list(missing))
+            LOG.warning(
+                "Requested unit(s) doesn't exist as %s: %s",
```
rohanpm, line 509:
```suggestion
                "Requested unit(s) don't exist as %s: %s",
```

tests/logs/delete/test_delete_packages/test_delete_files.txt
```diff
@@ -0,0 +1,23 @@
+[    INFO] Get files: started
+[ WARNING] Requested unit(s) doesn't exist as file: random.txt
+[    INFO] 3 unit(s) found for deletion
+[    INFO] Get files: finished
+[ WARNING] other.iso is not present in some-filerepo
+[    INFO] Deleting hello.iso from ['some-filerepo']
```
rohanpm, line 6: Can we please avoid default string formatting of list in all these logs, and instead join them?  So it'd say "Deleting hello.iso from repo1, repo2" instead of "Deleting hello.iso from ['repo1', 'repo2']".  The string formatting of a list seems developer-oriented.

rohanpm (changes requested):
This seems pretty close, maybe you want to take it out of Draft status soon.

pubtools/_pulp/tasks/delete.py
rajulkumar: Added option ```--skip```  and ```@step("Publish")``` exists in the ```Publisher``` class

pubtools/_pulp/tasks/delete.py
rajulkumar: Done

pubtools/_pulp/tasks/delete.py
rajulkumar: Done

tests/logs/delete/test_delete_packages/test_delete_files.txt
rajulkumar: Done

pubtools/_pulp/tasks/delete.py
rajulkumar: Added steps for specific content type like ```@step("Delete advisories")``` or ```@step(Delete RPMs"")``` instead of general ```@step("Delete content")``` as it provided more granular control on the steps.

negillett (approved):

pubtools/_pulp/tasks/delete.py
```diff
@@ -0,0 +1,737 @@
...
+                    unit_map.get(unit_name).repos.append(repo)
+
+        missing = set(repos) - set(repo_map.keys())
+        if missing:
+            missing = ", ".join(sorted(list(missing)))
+            LOG.warning("No units to remove from %s", missing)
+
+        return repo_map, unit_map
+
+    @step("Remove RPMs")
```
rohanpm, line 546: Minor - from the user's point of view, this is going to show both a "remove" and a "delete" step for each content type.  I guess users are not going to understand what's the difference without looking into the code, and it could be confusing.

Maybe this one doesn't need to be a step, or maybe it could use the more Pulp-specific technical term "Unassociate RPMs", so we'd still have the timing info but non-Pulp-expert users would just ignore it rather than being confused.

rohanpm (changes requested):
About the tests, none of them seem to verify that there has actually been a deletion (by checking the state of the fake pulp client afterward).  It means the remove_content calls could be potentially wrong or missing without the tests detecting it.

Can some tests please do that?  For example, use search_content on the fake pulp client after the command completes to verify that some content was really deleted, and/or use compare_extra (like in test_push.py) to compare the entire pulp state against a baseline during the test.

rohanpm (approved):

PR #123, opened 2022-03-27
Update product_versions on upload of productid

rohanpm:
The legacy push code implemented the following side-effect:
when a productid cert is uploaded into a repo, version strings are
extracted from the cert and appended to the product_versions field in
that repo plus other repos with certain matching notes.

That side-effect of productid upload remains important, but was missing
from pubtools-pulp-push. Add it now.

### DIFF
--- a/pubtools/_pulp/services/cachingpulp.py
+++ b/pubtools/_pulp/services/cachingpulp.py
@@ -34,6 +34,17 @@ class CachingPulpClient(object):
 
         return out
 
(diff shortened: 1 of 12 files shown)


rohanpm:
It'd be best for this to wait on https://github.com/release-engineering/pubtools-pulp/pull/124 and rebase afterward.

rajulkumar (approved):

PR #126, opened 2022-04-04
Garbage collect arc content

jajreidy:
Adds removing old items from the all-rpm-content repository.

### DIFF
--- a/pubtools/_pulp/tasks/garbage_collect.py
+++ b/pubtools/_pulp/tasks/garbage_collect.py
@@ -1,13 +1,14 @@
 import logging
 from datetime import datetime, timedelta
 
-from pubtools.pulplib import Criteria, Matcher
+from pubtools.pulplib import Criteria, Matcher, RpmUnit
 
 from pubtools._pulp.task import PulpTask
 from pubtools._pulp.services import PulpClientService
 
 
 LOG = logging.getLogger("pubtools.pulp")
+step = PulpTask.step
 
 
 class GarbageCollect(PulpClientService, PulpTask):
@@ -31,7 +32,19 @@ class GarbageCollect(PulpClientService, PulpTask):
             default=5,
         )
 
+        self.parser.add_argument(
+            "--arc-threshold",
+            help="delete all-rpm-content older than this many days",
+            type=int,
+            default=30,
+        )
+
     def run(self):
+        self.delete_temp_repos()
+        self.clean_all_rpm_content()
+
+    @step("Delete temporary repos")
+    def delete_temp_repos(self):
         LOG.debug("Garbage collection begins")
         criteria = Criteria.and_(
             Criteria.with_field("notes.created", Matcher.exists()),
@@ -61,6 +74,37 @@ class GarbageCollect(PulpClientService, PulpTask):
                 LOG.error(out.error_details or out.error_summary)
 
         LOG.info("Temporary repo(s) deletion completed")
+
+    @step("Clean all-rpm-content")
+    def clean_all_rpm_content(self):
+        # Clear out old all-rpm-content
+        LOG.info("Start old all-rpm-content deletion")
+        arc_threshold = self.args.arc_threshold
+        criteria = Criteria.and_(
+            Criteria.with_unit_type(RpmUnit),
+            Criteria.with_field(
+                "cdn_published",
+                Matcher.less_than(datetime.utcnow() - timedelta(days=arc_threshold)),
+            ),
+        )
+        clean_repos = list(
+            self.pulp_client.search_repository(
+                Criteria.with_field("id", "all-rpm-content")
+            )
+        )
+        if not clean_repos:
+            LOG.info("No repos found for cleaning.")
+            return
+        arc_repo = clean_repos[0]
+        deleted_arc = list(arc_repo.remove_content(criteria=criteria))
+        deleted_content = []
+        for task in deleted_arc:
+            if task.repo_id == "all-rpm-content":
+                for unit in task.units:
+                    LOG.info("Old all-rpm-content deleted: %s", unit.name)
+                    deleted_content.append(unit)
+        if not deleted_content:
+            LOG.info("No all-rpm-content found older than %s", arc_threshold)
 
 
 def entry_point():
--- a/tests/garbage_collect/test_garbage_collect.py
+++ b/tests/garbage_collect/test_garbage_collect.py
@@ -6,6 +6,8 @@ from more_executors.futures import f_return
 from pubtools.pulplib import (
     FakeController,
     Repository,
+    RpmUnit,
+    RpmDependency,
     YumRepository,
     Task,
     InvalidDataException,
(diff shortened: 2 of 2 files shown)


pubtools/_pulp/tasks/garbage_collect.py
```diff
@@ -31,6 +31,13 @@ def add_args(self):
             default=5,
         )
 
+        self.parser.add_argument(
+            "--arc-threshold",
+            help="delete repos older than this many days",
```
rohanpm, line 36: The help text is not accurate.

pubtools/_pulp/tasks/garbage_collect.py
```diff
@@ -62,6 +69,22 @@ def run(self):
 
         LOG.info("Temporary repo(s) deletion completed")
 
+        # Clear out old all-rpm-content
```
rohanpm, line 72: May I suggest putting this into a separate method rather than making run() bigger?

In fact the best would be also to refactor the existing code to end up with something like this:

```python
def run(self):
    self.delete_temp_repos()
    self.clean_all_rpm_content()

@step("Delete temporary repos")
def delete_temp_repos(self):
    ... # all the logic previously in run()

@step("Clean all-rpm-content")
def clean_all_rpm_content(self):
    ... # the new logic you're adding
```

The `@step` decorator puts some standard logging around sub-steps of a task.

pubtools/_pulp/tasks/garbage_collect.py
```diff
@@ -62,6 +69,22 @@ def run(self):
...
+            Criteria.with_field("cdn_published", Matcher.less_than(datetime.utcnow() - timedelta(days=arc_threshold))),
+        )
+        try:
+            arc_repo = self.pulp_client.get_repository("all-rpm-content").result()
+            deleted_arc = list(
+                arc_repo.remove_content(criteria=criteria)
+            )
+            LOG.info("Old all-rpm-content deletion completed %s", len(deleted_arc[0].units))
+        except PulpException:
+            LOG.info("all-rpm-content repo unavailable")
```
rohanpm, line 86: This isn't right, because you don't know if that's the reason for raising PulpException.  If it was raised for any other reason, it's not right to throw away the exception details and output a guess/lie in the logs.

How about changing it like this:

```python
clean_repos = list(self.pulp_client.search_repository(Criteria.with_field("id", "all-rpm-content")))
if not clean_repos:
    LOG.info("No repos found for cleaning.")
    return
arc_repo = clean_repos[0]
# ...then the rest
```

You would tolerate absence of all-rpm-content without silencing any other errors.

pubtools/_pulp/tasks/garbage_collect.py
```diff
@@ -62,6 +69,22 @@ def run(self):
...
+        criteria = Criteria.and_(
+            Criteria.with_unit_type(RpmUnit),
+            Criteria.with_field("cdn_published", Matcher.less_than(datetime.utcnow() - timedelta(days=arc_threshold))),
+        )
+        try:
+            arc_repo = self.pulp_client.get_repository("all-rpm-content").result()
+            deleted_arc = list(
+                arc_repo.remove_content(criteria=criteria)
+            )
+            LOG.info("Old all-rpm-content deletion completed %s", len(deleted_arc[0].units))
```
rohanpm, line 84: Please avoid writing the code in a way which assumes exactly one task is returned by remove_content.  You should loop over the tasks, even if it seems like there's always just one.

Also, can we please log the actual units and not just the number of units?  I'd find that info useful.

tests/garbage_collect/test_garbage_collect.py
```diff
@@ -151,3 +153,191 @@ def test_entry_point(mock_logger):
...
+    existing_rpm1 = RpmUnit(
+        cdn_published=datetime.datetime.utcnow(),
+        arch="src",
+        filename="test-arc01-1.0-1.src.rpm",
+        md5sum="ba9257ced24f77f4d777e399e67924f5",
+        name="test-arc01",
+        version="1.0",
+        release="1",
+        provides=[],
+        requires=[
```
rohanpm, line 189: You might consider leaving some of these fields out, particularly `requires`, just so the test doesn't need as much code.  These fields are irrelevant to this specific test.

tests/garbage_collect/test_garbage_collect.py
```diff
@@ -151,3 +153,191 @@ def test_entry_point(mock_logger):
...
+    arg = ["", "--pulp-url", "http://some.url", "--arc-threshold", "7"]
+
+    with patch("sys.argv", arg):
+        gc_args = gc.args
+
+    assert hasattr(gc_args, "arc_threshold")
+    assert gc_args.arc_threshold == 7
+
+def test_arc_garbage_collect(mock_logger):
+    """deletes the arc repo that confirms to garbage collect criteria"""
```
rohanpm, line 170: It doesn't delete the arc repo, it deletes content from the repo.

rohanpm (changes requested):

pubtools/_pulp/tasks/garbage_collect.py
```diff
@@ -62,6 +75,38 @@ def run(self):
...
+            LOG.info("No repos found for cleaning.")
+            return
+        arc_repo = clean_repos[0]
+        deleted_arc = list(arc_repo.remove_content(criteria=criteria))
+        deleted_content = []
+        for task in deleted_arc:
+            if task.repo_id == "all-rpm-content":
+                for unit in task.units:
+                    LOG.info("Old all-rpm-content deleted: %s", unit.name)
+                    print("Old all-rpm-content deleted: %s", unit.name)
```
rohanpm, line 105: Can you please drop the print?

rohanpm (changes requested):
Could you please rebase or squash this PR so we get just one nice commit before merge, without all the intermediate versions of this work?

pubtools/_pulp/tasks/garbage_collect.py
```diff
@@ -31,7 +32,19 @@ def add_args(self):
             default=5,
         )
 
+        self.parser.add_argument(
+            "--arc-threshold",
+            help="delete all-rpm-content older than this many days",
+            type=int,
+            default=180,
```
rohanpm, line 39: Hmm, I think we should make this default a fair bit more aggressive.  How about 30?

An older variant of the cleanup code, rhsm-pulp-clean-rpms, used a default of 180 and it's probably been carried over from there.  However, the old code worked differently.  It used the timestamp of "file arrived in pulp" while the new code uses a different timestamp, cdn_published, meaning "file was published from Pulp to CDN".  I think it's fine to use that timestamp, but it means this can be quite more aggressive as we already know that the file must have been published from Pulp to CDN by the time we're cleaning it up.

rohanpm:
Could you please squash/rebase the PR to a single commit, e.g. using `git rebase -i` ?

rohanpm:
@jajreidy Could you please squash/rebase the PR to a single commit, e.g. using git rebase -i ?

CHANGELOG.md
```diff
@@ -128,9 +128,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Initial release to PyPI
 
+<<<<<<< HEAD
```
rohanpm, line 131: @jajreidy you might have had some trouble with squashing commits.  Markers like this shouldn't be here, and also the PR still contains 18 commits.  Could you have another try?  I can show you on a screen share if you want.

rohanpm (changes requested):

rohanpm:
This has all been squashed into a single commit now.  @jajreidy can you please confirm here that the final content of the PR looks correct before merging?

jajreidy:
> This has all been squashed into a single commit now. @jajreidy can you please confirm here that the final content of the PR looks correct before merging?

Checked through everything and it seems to match what I have locally. So should be good.

rohanpm (approved):

PR #143, opened 2022-05-29
refactor: split progress logging from context

rohanpm:
The progress logging does not have much to do with the rest of the
Context class. Move it elsewhere so that each class is more focused on a
single responsibility.


rohanpm:
Abandoning this as there is a larger PR incorporating the same changes plus more.

PR #146, opened 2022-06-08
Added PostPushActions phase

midnightercz:
New phase should serve as phase to do any neccesary post push actions.
Currently being implemented to trigger new hook pulp_item_finished
which is be called with pushed pulp items
To improve static score order of imports in some modules was changed

### DIFF
--- a/pubtools/_pulp/hooks.py
+++ b/pubtools/_pulp/hooks.py
@@ -14,4 +14,13 @@ def task_pulp_flush():
     Pulp content may have recently changed."""
 
 
+@hookspec
+def pulp_item_finished(item_metadata=None): # pylint: disable=unused-argument
+
+    """Invoked when PulpPushItem is pushed to a pulp target
+
+    This hook should be invoked per every pub push item after end of push
+    process."""
+
+
 pm.add_hookspecs(sys.modules[__name__])
--- a/pubtools/_pulp/hooks.py
+++ b/pubtools/_pulp/hooks.py
@@ -15,14 +15,22 @@ def task_pulp_flush():
 
 
 @hookspec
-def pulp_item_finished(
-    item_metadata=None, push_item=None
+def pulp_item_push_finished(
+    pulp_units=None, push_item=None
 ):  # pylint: disable=unused-argument
 
-    """Invoked when PulpPushItem is pushed to a pulp target
-
-    This hook should be invoked per every pub push item after end of push
-    process."""
+    """Invoked during push tasks after each item has been processed fully.
+    
+    By the time this hook is invoked, the referenced item and unit is expected
+    to be fully uploaded into Pulp and published onto the CDN.
+    Args:
+        pulp_units (list[:class:`~pubtools.pulplib.Unit`])
+            A list of zero or more Pulp unit(s) created/updated for this item.
+            Note that this information may not be available for every content type,
+            and may only contain a subset of the Pulp fields.
+        push_item (:class:`~pushsource.PushItem`)
+            The item which has been pushed.
+    """
 
 
 pm.add_hookspecs(sys.modules[__name__])
--- a/pubtools/_pulp/hooks.py
+++ b/pubtools/_pulp/hooks.py
@@ -23,6 +23,7 @@ def pulp_item_push_finished(
 
     By the time this hook is invoked, the referenced item and unit is expected
     to be fully uploaded into Pulp and published onto the CDN.
+
     Args:
         pulp_units (list[:class:`~pubtools.pulplib.Unit`])
             A list of zero or more Pulp unit(s) created/updated for this item.
--- a/pubtools/_pulp/hooks.py
+++ b/pubtools/_pulp/hooks.py
@@ -15,7 +15,7 @@ def task_pulp_flush():
 
 
 @hookspec
-def pulp_item_finished(item_metadata=None):  # pylint: disable=unused-argument
+def pulp_item_finished(item_metadata=None, push_item=None):  # pylint: disable=unused-argument
 
     """Invoked when PulpPushItem is pushed to a pulp target
 
--- a/pubtools/_pulp/hooks.py
+++ b/pubtools/_pulp/hooks.py
@@ -15,7 +15,7 @@ def task_pulp_flush():
 
 
 @hookspec
-def pulp_item_finished(item_metadata=None): # pylint: disable=unused-argument
+def pulp_item_finished(item_metadata=None):  # pylint: disable=unused-argument
 
     """Invoked when PulpPushItem is pushed to a pulp target
 
--- a/pubtools/_pulp/tasks/push/command.py
+++ b/pubtools/_pulp/tasks/push/command.py
@@ -148,8 +148,8 @@ class Push(
             add_phase(Associate)
 
             if "publish" in self.args.skip:
-                # Caller doesn't want to publish, then we just do post push actions
-                # for push items and wait for prior phases to complete.
+                # Caller doesn't want to publish, then we just wait for prior phases
+                # to complete
                 add_phase(EndPush)
 
             else:
--- a/pubtools/_pulp/tasks/push/command.py
+++ b/pubtools/_pulp/tasks/push/command.py
@@ -145,7 +145,7 @@ class Push(
             add_phase(Update)
 
             # Ensure all items are associated into the desired target repos.
-            associate_phase = add_phase(Associate)
+            add_phase(Associate)
 
             if "publish" in self.args.skip:
                 # Caller doesn't want to publish, then we just do post push actions
--- a/pubtools/_pulp/tasks/push/command.py
+++ b/pubtools/_pulp/tasks/push/command.py
@@ -113,9 +113,7 @@ class Push(
                 update_push_items=collect_phase.in_queue.put,
                 publish_with_cache_flush=self.publish_with_cache_flush,
             )
-            phase_inst = klass(**kwargs)
-            phases.append(phase_inst)
-            return phase_inst
+            phases.append(klass(**kwargs))
 
         # Now proceed with adding the phases which make up a push...
 
@@ -139,7 +137,6 @@ class Push(
         if self.args.pre_push:
             # If we are in pre-push mode then we do not go any further, we just wait
             # for all previous steps, do post actions for push items, then log a message and exit.
-            add_phase(PostPushActions)
             add_phase(EndPush)
 
         else:
@@ -153,7 +150,6 @@ class Push(
             if "publish" in self.args.skip:
                 # Caller doesn't want to publish, then we just do post push actions
                 # for push items and wait for prior phases to complete.
-                add_phase(PostPushActions)
                 add_phase(EndPush)
 
             else:
--- a/pubtools/_pulp/tasks/push/command.py
+++ b/pubtools/_pulp/tasks/push/command.py
@@ -15,6 +15,7 @@ from .phase import (
     Publish,
     Context,
     ProgressLogger,
+    PostPushActions,
 )
 from ..common import Publisher, PulpTask
 from ...services import (
@@ -112,7 +113,9 @@ class Push(
                 update_push_items=collect_phase.in_queue.put,
                 publish_with_cache_flush=self.publish_with_cache_flush,
             )
-            phases.append(klass(**kwargs))
+            phase_inst = klass(**kwargs)
+            phases.append(phase_inst)
+            return phase_inst
 
         # Now proceed with adding the phases which make up a push...
 
@@ -135,7 +138,8 @@ class Push(
 
         if self.args.pre_push:
             # If we are in pre-push mode then we do not go any further, we just wait
-            # for all previous steps, then log a message and exit.
+            # for all previous steps, do post actions for push items, then log a message and exit.
+            add_phase(PostPushActions)
             add_phase(EndPush)
 
         else:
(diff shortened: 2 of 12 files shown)


pubtools/_pulp/tasks/push/command.py
```diff
@@ -15,6 +15,7 @@
     Publish,
     Context,
     ProgressLogger,
+    PostPushActions,
```
rohanpm, line 18: I think you forgot to "git add" the file with PostPushActions.

pubtools/_pulp/hooks.py
```diff
@@ -14,4 +14,13 @@ def task_pulp_flush():
     Pulp content may have recently changed."""
 
 
+@hookspec
+def pulp_item_finished(item_metadata=None):  # pylint: disable=unused-argument
```
rohanpm, line 18: The PostPushActions code is missing right now, but based on context I see what this is going for.

I've got some issues with the data currently being passed to this hook: the PulpPushItem classes are meant to be internal, they are not designed as public API.  But as soon as we start passing them into this hook then we are "locked in", we can't refactor them any more without breaking whoever's on the other end of the hook.  We need to be strict about what we're producing for this hook in order to avoid issues.  Some of the fields you're exporting now (like upload_key) were absolutely intended to be just "this is what works for the code internally" and it is very risky to let other libraries start depending on them.

Is it possible to satisfy your use-case by yielding only public classes from pushsource and/or pubtools-pulplib libraries?  I'm guessing that it might be - you're interested in cdn_path but that's a documented field on FileUnit at https://release-engineering.github.io/pubtools-pulplib/api/files.html#pubtools.pulplib.FileUnit.cdn_path .

What do you think of this alternative signature then:

```python
def pulp_item_finished(pushitem, pulp_units):
```

...where:

- pushitem is an instance of [pushsource.PushItem](https://release-engineering.github.io/pushsource/model/base.html#pushsource.PushItem)
- pulp_units is a list of [pulplib.Unit](https://release-engineering.github.io/pubtools-pulplib/api/pulpcore.html#pubtools.pulplib.Unit) (list keeps the possibility open of one PushItem generating zero units or more than one unit).  These are the units created or updated in Pulp for that item.
- Obviously it would be the concrete subclasses like FilePushItem, FileUnit depending on the specific content type.

tests/push/test_push.py
```diff
@@ -129,15 +129,18 @@ def test_typical_push(
...
-    assert len(hookspy) == 11
+    assert len(hookspy) == 24
     (hook_name, hook_kwargs) = hookspy[0]
     assert hook_name == "task_start"
     (hook_name, hook_kwargs) = hookspy[1]
     assert hook_name == "pulp_repository_pre_publish"
     (hook_name, hook_kwargs) = hookspy[2]
     assert hook_name == "pulp_repository_published"
-    (hook_name, hook_kwargs) = hookspy[-2]
+    # after pulp_repository_published there's 13 calls of pulp_item_finished
```
rohanpm, line 139: There should be at least one test which verifies the actual data passed into the hook, otherwise it can change without warning and break anything connected to the hook.

rohanpm (changes requested):

pubtools/_pulp/hooks.py
midnightercz: yes, I was particularly interested in cdn_path

pubtools/_pulp/tasks/push/phase/push_post_actions.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, context, in_queue, **kwargs):
+        super(PostPushActions, self).__init__(
+            context, in_queue=in_queue, name="Post Push Actions", **kwargs
+        )
+
+    def run(self):
+        for item_batch in self.iter_input_batched():
+            for item in item_batch:
+                pm.hook.pulp_item_finished(
+                    item_metadata=attrs.asdict(item.pulp_unit)
```
rohanpm, line 36: Could you please send the actual unit rather than converting to dict?  The unit is a properly documented class with stable API.  If you would prefer it as dict you can always convert it on the other side of the hook.

pubtools/_pulp/hooks.py
```diff
@@ -14,4 +14,15 @@ def task_pulp_flush():
...
 
+@hookspec
+def pulp_item_finished(
+    item_metadata=None, push_item=None
+):  # pylint: disable=unused-argument
+
+    """Invoked when PulpPushItem is pushed to a pulp target
+
+    This hook should be invoked per every pub push item after end of push
+    process."""
```
rohanpm, line 25: - The doc should be fleshed out a bit I think, also rewritten because PulpPushItem is not a thing which makes sense anywhere outside of this code.
- Concrete types should probably be documented as well.
- Later in the change I'm suggesting to yield actual unit and not a dict, so the name of that argument should change.
- As suggested I think this should have a list of units, not a single unit, to reserve the possibility of an item generating multiple units.  modulemd and comps.xml are both current examples of push items which end up creating multiple units (though not tracked by pubtools-pulp today).
- And finally we only invoke this hook for push tasks, and not other kinds of tasks which can also process "items" (like publish or fix CVEs), so the hook name should probably reflect that.

So overall I suggest following changes:

```suggestion
def pulp_item_push_finished(
    pulp_units=None, push_item=None
):  # pylint: disable=unused-argument
    """Invoked during push tasks after each item has been processed fully.
    
    By the time this hook is invoked, the referenced item and unit is expected
    to be fully uploaded into Pulp and published onto the CDN.

    Args:
        pulp_units (list[:class:`~pubtools.pulplib.Unit`])
            A list of zero or more Pulp unit(s) created/updated for this item.

            Note that this information may not be available for every content type,
            and may only contain a subset of the Pulp fields.

        push_item (:class:`~pushsource.PushItem`)
            The item which has been pushed.
    """
```

pubtools/_pulp/tasks/push/command.py
```diff
@@ -112,7 +113,9 @@ def add_phase(klass, **kwargs):
                 update_push_items=collect_phase.in_queue.put,
                 publish_with_cache_flush=self.publish_with_cache_flush,
             )
-            phases.append(klass(**kwargs))
+            phase_inst = klass(**kwargs)
+            phases.append(phase_inst)
+            return phase_inst
```
rohanpm, line 118: This change and the later `associate` assignment seem to be unused, maybe from some earlier work in progress?  Could you take them out please?

pubtools/_pulp/tasks/push/command.py
```diff
@@ -135,7 +138,8 @@ def add_phase(klass, **kwargs):
 
         if self.args.pre_push:
             # If we are in pre-push mode then we do not go any further, we just wait
-            # for all previous steps, then log a message and exit.
+            # for all previous steps, do post actions for push items, then log a message and exit.
+            add_phase(PostPushActions)
```
rohanpm, line 142: I'm not sure this should be invoked for the pre-push case.  Given that we are saying the hook is invoked when an item is pushed, and pre-push doesn't really push them (make things available on the CDN), it doesn't seem to fit.

If it is necessary for your use-case - how do you plan to differentiate pushed vs non-pushed items?  If you don't plan to use it in the pre-push case I'd rather not add it.

pubtools/_pulp/tasks/push/command.py
```diff
@@ -144,16 +148,19 @@ def add_phase(klass, **kwargs):
...
             # Ensure all items are associated into the desired target repos.
-            add_phase(Associate)
+            associate_phase = add_phase(Associate)
 
             if "publish" in self.args.skip:
-                # Caller doesn't want to publish, then we just wait for prior phases
-                # to complete.
+                # Caller doesn't want to publish, then we just do post push actions
+                # for push items and wait for prior phases to complete.
+                add_phase(PostPushActions)
```
rohanpm, line 156: Similarly to the pre-push case, are you sure you want to invoke the hook if publish didn't happen?  Or do you want to add an argument onto the hook which says whether or not publish happened, so the other side can tell?

pubtools/_pulp/tasks/push/command.py
```diff
@@ -144,16 +148,19 @@ def add_phase(klass, **kwargs):
...
                 add_phase(EndPush)
 
             else:
-                # Ensure all repos are published once the desired content is present.
+                # Ensure all repos are published once the desired content is present
+                # and do any post push pushitems actions.
                 add_phase(Publish)
+                add_phase(PostPushActions)
```
rohanpm, line 163: This is prone to hanging because PostPushActions has an output queue, but there is no later phase reading from it.  If the output queue fills up then the phase is going to deadlock.

For correct behavior it's required that the very last phase has no output queue.  My suggestion is that you remove the output queue from PostPushActions, and in the case above where it goes (PostPushActions => EndPush), you can just remove EndPush and leave PostPushActions as the last phase.

tests/push/test_push.py
rohanpm: ...and by actual data I mean something more precise than mock.ANY.

rohanpm (changes requested):
No major issues with this and I think it's pretty close to done, however I ask again can you please share a bit more info about the use-case and what this is supposed to accomplish?

It seems like you need to do something during Pulp pushes, but in the epic CLOUDDST-10142 there's no mention of this I can find.  In fact there's at least one issue which points at something else: CLOUDDST-12516 talks about creating a target just for CG.  So is CG going to be used from both pulp targets and CG targets or what?  Is there an up-to-date doc which explains more about it?

tests/push/test_push.py
midnightercz: Hmm, but that's kinda changing unit test to integration test isn't it? When hook is called with attrs.asdict(item.pulp_unit). When you add or remove any attributes in pulp unit tests will fail.

pubtools/_pulp/tasks/push/phase/push_post_actions.py
midnightercz: Yeah, I can do that. I exported dicts as it's more memory efficient as push items can be freed after pulp push finishes.

pubtools/_pulp/tasks/push/command.py
midnightercz: Ok, I added this only to support pre-push type of task. It's not needed for my use case

tests/push/test_push.py
rohanpm: I'm not interested in trying to label each test as a certain type (unit/integration/system). I'm more interested in which bugs the test suite can detect.

In this case if the arguments weren't tested at all or were tested as mock.ANY, then even critical and obvious bugs such as calling the hook with swapped arguments would still pass all tests.

I think the way you've updated it now, where the arguments are tested but just lightly by only looking at some of the fields or types, is a good middle ground.  As it's quite unlikely there could be a bug where the hook is invoked with an object which is simultaneously a PushItem, but the wrong PushItem.

So I think the approach right now is decent, but it will need some changes to align with the other suggestions about yielding actual PulpUnit instances rather than dicts.

pubtools/_pulp/hooks.py
```diff
@@ -14,4 +14,23 @@ def task_pulp_flush():
...
+    pulp_units=None, push_item=None
+):  # pylint: disable=unused-argument
+
+    """Invoked during push tasks after each item has been processed fully.
+
+    By the time this hook is invoked, the referenced item and unit is expected
+    to be fully uploaded into Pulp and published onto the CDN.
+    Args:
+        pulp_units (list[:class:`~pubtools.pulplib.Unit`])
+            A list of zero or more Pulp unit(s) created/updated for this item.
```
rohanpm, line 28: There is currently a mismatch between the docs and what you implemented.

Docs say that it's a list of zero or more, and doesn't say anything about it possibly being None.  This implies that if there's no units, it'll give you an empty list.

What you've implemented instead is that pulp_units will be None if there were no units.

Could you please make the docs and implementation match?  Probably by yielding an empty list rather than None in the 'no units' case.

rohanpm (changes requested):

pubtools/_pulp/tasks/push/command.py
```diff
@@ -147,13 +148,15 @@ def add_phase(klass, **kwargs):
             add_phase(Associate)
 
             if "publish" in self.args.skip:
-                # Caller doesn't want to publish, then we just wait for prior phases
-                # to complete.
+                # Caller doesn't want to publish, then we just do post push actions
```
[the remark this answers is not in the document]
midnightercz: right, I forgot to revert this

pubtools/_pulp/tasks/push/phase/publish.py
```diff
@@ -105,5 +101,6 @@ def run(self):
         # Note we don't keep track of exactly which items got published through each
         # repo, so this will simply show that everything moved from in progress to done
         # at once.
-        for _ in pushed_items:
+        for item in pushed_items:
             self.progress_info.incr_out()
```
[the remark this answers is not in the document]
midnightercz: ah, I see. So I can just remove explicit call of incr_out

pubtools/_pulp/tasks/push/command.py
```diff
@@ -135,7 +136,7 @@ def add_phase(klass, **kwargs):
 
         if self.args.pre_push:
             # If we are in pre-push mode then we do not go any further, we just wait
-            # for all previous steps, then log a message and exit.
+            # for all previous steps, do post actions for push items, then log a message and exit.
```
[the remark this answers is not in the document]
rohanpm: This might be the last thing to fix.

rohanpm (approved):

PR #154, opened 2022-07-25
Increase retry backoff RHELDST-9957

amcmahon-rh:
Several tasks have failed to flush the UD Cache, for both repo and errata data types. Previous investigations pointed to an issue with the /internal/rcm/flush-cache endpoint, which is an API managed by teamnado. Teamnado folks have said the host running this endpoint has had some issues. Between this info, the fact that the task can succeed when retried later and the exceptions that occur, I believe the intermittent host unavailability is the root issue.

This PR duplicates the same changes I made in pub-utils, increasing the total duration over which retries are carried out. If the root cause of failures is due to intermittent outages, extending the waiting time between retries should allow the issue to pass.


amcmahon-rh:
For some reason the rpm-py-installer package no longer works with python3.8. I'm looking into it atm

tox.ini
```diff
@@ -14,6 +14,14 @@ deps=
 	-clegacy.constraints
 	-rtest-requirements.in
 
+[testenv:py38]
```
rohanpm, line 17: I guess you've added this for debugging the CI failure relating to rpm-py-installer.  Could you drop it now please?  That problem has disappeared.

rohanpm (changes requested):

rohanpm (approved):

rajulkumar (approved):

PR #176, opened 2022-11-11
Fixed pulp_item_push_finished hook arguments

midnightercz:


pubtools/_pulp/hooks.py
```diff
@@ -16,7 +16,7 @@ def task_pulp_flush():
 
 @hookspec
 def pulp_item_push_finished(
-    pulp_units=None, push_item=None
+    pulp_units, push_item
```
rohanpm, line 19: Please take all the changes other than this one out of your commit/PR.

rohanpm (changes requested):

rohanpm (changes requested):
Could you please fix up the commit message?

- It now has some incorrect info about gssapi
- It would be useful if it summarized exactly *why* this is a fix.  Why was it wrong to have default arguments of `None`?  Why is it right to remove them?

rohanpm (changes requested):
Could you please fix up the commit message?

- It now has some incorrect info about gssapi
- It would be useful if it summarized exactly why this is a fix. Why was it wrong to have default arguments of None? Why is it right to remove them?

rohanpm (approved):

PR #189, opened 2023-01-10
Add PKI support for pulp client

alexverdes:
With the move to certificates and read-only passwords based authentication, the following were implemented:
- adds --pulp-certificate and --pulp-certificate-key arguments
- supports single file (e.g. .pem) or multiple files (e.g. .cert,.key)  certificates
- supports looking up the cert/key path via the get_cert_key_paths hook defined here: https://release-engineering.github.io/pubtools/hooks.html

### DIFF
--- a/pubtools/_pulp/services/pulp.py
+++ b/pubtools/_pulp/services/pulp.py
@@ -5,6 +5,7 @@ import logging
 import warnings
 
 from pubtools import pulplib
+from pubtools.pluggy import pm
 
 from .base import Service
 from .fakepulp import new_fake_controller
@@ -48,6 +49,16 @@ class PulpClientService(Service):
             default=None,
         )
         group.add_argument(
+            "--pulp-certificate",
+            help="Pulp certificate. Can also be a single file (.pem)",
+            default=None,
+        )
+        group.add_argument(
+            "--pulp-certificate-key",
+            help="Pulp certificate key",
+            default=None,
+        )
+        group.add_argument(
             "--pulp-insecure",
             action="store_true",
             help="Allow unverified HTTPS connection to Pulp",
@@ -87,8 +98,12 @@ class PulpClientService(Service):
 
     def new_pulp_client(self, **kwargs):
         """Creates and returns a new Pulp client with appropriate config."""
+        cert = None
         auth = None
         args = self._service_args
+
+        # Use hook to get the certificate path(s). Defined here, used later.
+        hook_rets = pm.hook.get_cert_key_paths(server_url=args.pulp_url)
 
         if not args.pulp_fake and not args.pulp_url:
             LOG.error("At least one of --pulp-url or --pulp-fake must be provided")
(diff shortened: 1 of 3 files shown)


tests/push/test_push.py
```diff
@@ -131,13 +131,13 @@ def test_typical_push(
     )
 
     # It should have invoked hook(s).
-    assert len(hookspy) == 24
+    assert len(hookspy) == 25
     (hook_name, hook_kwargs) = hookspy[0]
     assert hook_name == "task_start"
     (hook_name, hook_kwargs) = hookspy[1]
-    assert hook_name == "pulp_repository_pre_publish"
```
rohanpm, line 138: Instead of changing the existing hook_name comparisons, could we please insert a new comparison?  i.e., before we tested hookspy 0 through 2, now let's test hookspy 0 through 3, so we still test everything we tested before but also the one new hook invocation.

Otherwise there is no testing of the invocation of pulp_repository_published hook (which is referenced in a comment on line 141, which would no longer make sense).

pubtools/_pulp/services/pulp.py
```diff
@@ -47,6 +49,17 @@ def add_service_args(self, parser):
             help="Pulp password (or set PULP_PASSWORD environment variable)",
             default=None,
         )
+        group.add_argument(
+            "--pulp-certificate",
+            help="Pulp certificate (or taken from '~/certs/$USER.crt'). "
```
rohanpm, line 54: The default doesn't seem to be implemented actually, or at least I can't find the code for it (and can't find any code using DEFAULT_CERT_PATH).

rohanpm (changes requested):

pubtools/_pulp/services/pulp.py
```diff
@@ -87,9 +98,13 @@ def pulp_fake_controller(self):
 
     def new_pulp_client(self, **kwargs):
         """Creates and returns a new Pulp client with appropriate config."""
+        cert = None
         auth = None
         args = self._service_args
 
+        # Use hook to get the certificate path(s). Defined here, used later.
+        hook_rets = pm.hook.get_cert_key_paths(server_url=args.pulp_url)
```
rohanpm, line 106: I looked at the implementation of this hook (in Pub) and it is unfortunately implemented to *raise an exception*, rather than return None, if no cert/key has been deployed.

That's going to cause a problem then, because we haven't deployed certs/keys yet (since we haven't implemented support for them in Pulp yet).  So as soon as this code is deployed onto Pub it'll start crashing.

We need to handle that somehow, and it should be handled before this change is submitted (otherwise pubtools-pulp becomes effectively unreleasable).

Given that the hook documentation at https://release-engineering.github.io/pubtools/hooks.html#get_cert_key_paths says that "Callers of this hook should be prepared to receive no result", I'd say it's the implementation in Pub which is wrong, as it will never return no result (you will always either get a client cert & key, or get an exception).

Would you be able to resolve that?  Either changing the hook in Pub so that it logs a warning and returns None instead of raising; or ensuring that we actually deploy some certs at the expected paths.

rohanpm (changes requested):

rohanpm:
Could this please be rebased now, ideally so that the PR contains only a single commit?

At the moment it shows unrelated changes which have happened in the repo since the PR was originally created, such as the dropping of python2 support.

alexverdes:
> Could this please be rebased now, ideally so that the PR contains only a single commit?
> 
> At the moment it shows unrelated changes which have happened in the repo since the PR was originally created, such as the dropping of python2 support.

Guessing this would be the expected result?

rohanpm:
> > Could this please be rebased now, ideally so that the PR contains only a single commit?
> > At the moment it shows unrelated changes which have happened in the repo since the PR was originally created, such as the dropping of python2 support.
> 
> Guessing this would be the expected result?

If you mean the result after your push about 1 hour ago, yes it looks good now.

rohanpm (approved):

PR #196, opened 2023-02-27
Added container_list converter for ErratumUnit

midnightercz:


rohanpm:
What's here looks fine but this needs to include a bump in the minimum required version of pushsource and pubtools-pulplib.

midnightercz:
> What's here looks fine but this needs to include a bump in the minimum required version of pushsource and pubtools-pulplib.
Yes, but pushsource needs to be released first. I'll do that once pushsource is updated

rbikar (approved):

PR #201, opened 2023-03-20
UD cache flush for Erratum in CVE fix task [RHELDST-16262]

rbikar:
Added UD cache flush of Erratum object in CVE fix task. With this fix Erratum related to given CVE should be available on Customer portal pages without delay.

### DIFF
--- a/tests/fix_cves/test_fix_cves.py
+++ b/tests/fix_cves/test_fix_cves.py
@@ -212,11 +212,12 @@ def test_fix_cves_with_cache_cleanup(command_tester):
 
         fastpurge_client = fake_fix_cves.fastpurge_client
 
-        assert fastpurge_client.purged_urls == ["https://cdn.example.com/content/unit/1/all-rpm/mutable1", 
-                                                "https://cdn.example.com/content/unit/1/all-rpm/mutable2",
-                                                "https://cdn.example.com/content/unit/1/client/mutable1",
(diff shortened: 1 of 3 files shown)


tests/fix_cves/test_fix_cves.py
```diff
@@ -21,6 +21,7 @@ class FakeUdCache(object):
     def __init__(self):
         self.flushed_repos = []
         self.flushed_products = []
+        self.flushed_errata = []
```
rohanpm, line 24: Minor: why add this and then not use it?

It seems like the existing test_fix_cves_with_cache_cleanup test probably ought to look at flushed_errata and verify the expected erratum was flushed.  You are after all making changes to this logic, and given that no other test updates were included here, it doesn't seem like the change is being tested beyond "it doesn't crash".  So nothing would prevent the same bug you're trying to fix from coming back later.

(The test really should have already been checking some of these attributes.  It seems like the class was copied from some other files where tests *do* use the attributes during verification, but such verification was never done for fix-cves, making all these attributes currently unused.  I'm not keen on adding another unused attribute just to be consistent with the earlier copy-pasted code...)

rohanpm (dismissed):

tests/fix_cves/test_fix_cves.py
rbikar: good point, will add a couple of asserts there.

rohanpm (approved):

PR #202, opened 2023-03-23
Add repo-lock to set-maintenance tasks

amcmahon-rh:
A repo lock feature has just been added to pubtools-pulplib which allows tasks to check to see if other tasks are modifying the repo. This feature was introduced because over a race condition found when setting maintenance status.

This change introduces the repo-locking to the set maintenance report tasks so that release engineers can run multiple set maintenance tasks at once and not worry about data being overwritten.


tests/set_maintenance/test_set_maintenance.py
```diff
@@ -41,6 +41,16 @@ def pulp_client(self):
         return self.pulp_client_controller.client
 
 
+@pytest.fixture(autouse=True)
+def constant_lock_id(monkeypatch):
+    # Locks are usually assigned a random ID to make them unique. For testing
+    # we want to keep this constant to compare logs easier.
+    monkeypatch.setattr(
+        "pubtools.pulplib._impl.model.repository.repo_lock.random.choices",
```
rohanpm, line 49: Please don't do it this way because it ties the pubtools-pulp test too closely to specific implementation details of pubtools-pulplib.

Even completely innocent refactors in pubtools-pulplib such as renaming repo_lock.py to something else, or changing the code from `import random` to `from random import choices`, would cause this test to start failing.

I suggest rather the following solution: in `CommandTester._normalize_plaintext`, there is already code for a similar case of having to normalize a randomly-generated tmpdir path into a placeholder.  Extend that so it replaces strings of the form `(Submitting|Deleting) lock with id '([^']+)'` with `\1 lock with id '<lock_id>'`.
Then we don't depend on how exactly the lock ID is generated, we only depend on how the log messages look, which is considered less of an implementation detail given that it's visible to end-users.

tests/set_maintenance/test_set_maintenance.py
```diff
@@ -41,6 +41,16 @@ def pulp_client(self):
         return self.pulp_client_controller.client
```
rohanpm, line 43: Could we please have at least one of the tests in here have a look at [repo_lock_history](https://release-engineering.github.io/pubtools-pulplib/api/testing.html#pubtools.pulplib.FakeController.repo_lock_history) to confirm that a lock and unlock happened?

rohanpm (changes requested):

rohanpm (approved):

rohanpm (commented):
Hmm, I should have checked the commits more closely before submit.

I'm not going to do anything about it now, but it looks like the git history in this PR is really messed up.  All the changes are attributed to commit 0ed7a034b3e9777661f6fc61f441da4463c93e4a, which is "Release 1.23.0" from Robert.  The metadata and message for that commit are a copy of the real "Release 1.23.0" from a15588f906af.

It seems like there has been a "git commit --amend", "git rebase" or similar which did not go as intended and the changes have been associated with the wrong author and with the wrong commit message.

PR #207, opened 2023-04-13
Fallback to defualt TTL [TODO]

rbikar:
TODO


pubtools/_pulp/cdn.py
```diff
@@ -142,3 +147,26 @@ def _format_template(ttl, template, path):
                 out.append(_format_template(ttl_ft, item, path))
 
         return out
+
+
+# ordering of items matters as it's used as priority
+CDN_TTL_CONFIG = OrderedDict(
```
rohanpm, line 153: Minor:

In my view a list of tuples is the more natural and better choice.  You never look up any items here by key, which is the main reason you'd want a dict in the first place.

Also, OrderedDict is mostly redundant from python 3.7 onwards as plain dicts are guaranteed to retain order, so this can be considered as creating some tech debt (adding some code which only needs to exist for python<3.7 and ideally would be cleaned up later).

rohanpm (approved):

PR #212, opened 2023-04-24
Published-before should accept datetime [RHELDST-12600]

amcmahon-rh:
Pub allows users to provide a datetime value for the published-before argument in the publish command, but the library implementation didn't accept it. This change adds in the extra format.

### DIFF
--- a/tests/publish/test_publish.py
+++ b/tests/publish/test_publish.py
@@ -385,10 +385,12 @@ def test_publish_repos_published_before_exception(command_tester):
                     "--published-before",
                     "2019-09-07BADFORMAT01:00:00Z",
                 ],
-                allow_raise=True
+                allow_raise=True,
             )
-        assert "published-before date should be in YYYY-mm-ddTHH:MM:SSZ " \
+        assert (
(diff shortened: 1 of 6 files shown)


tests/publish/test_publish.py
```diff
@@ -327,6 +330,69 @@ def test_publish_repos_published_before_a_date(command_tester):
...
+    # No repo should be published
+    assert [hist.repository.id for hist in fake_pulp.publish_history] == []
+
+
+def test_publish_repos_published_before_exception(command_tester):
+    """Expect a parser exception when passing a bad date"""
+    with pytest.raises(SystemExit) as e:
+        with FakePublish() as fake_publish:
+            fake_pulp = fake_publish.pulp_client_controller
+            _add_repo(fake_pulp)
```
rohanpm, line 380: You probably don't even need these lines, right?  As you expect the command to fail before it gets to the point of having any interactions with pulp.

rohanpm (approved):

rohanpm (approved):

PR #226, opened 2023-07-13
Retry all failed associations with up-to-date data

jm-wk:
During the association phase of push we first list the origin repos and then a while later we attempt to copy the items from origin repos to all other desired repos. Since the items can meanwhile be moved between repos, the association can fail on missing items.

We fix this by retrying the copy operation with up-to-date data from pulp.

### DIFF
--- a/pubtools/_pulp/tasks/push/items/base.py
+++ b/pubtools/_pulp/tasks/push/items/base.py
@@ -191,7 +191,9 @@ class PulpPushItem(object):
         return f_map(units_f, matcher)
 
     @classmethod
-    def associated_items_single_batch(cls, pulp_client, items, copy_options):
+    def associated_items_single_batch(
+        cls, pulp_client, items, copy_options, association_retry_count
+    ):
         """Associate a single batch of items into destination repos.
 
         This generator yields instances of Future[list[<associated-items>]].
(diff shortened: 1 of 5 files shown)


pubtools/_pulp/tasks/push/items/base.py
```diff
@@ -281,11 +298,30 @@ def refresh_after_copy(_):
...
+
+            # Return a future with list of all results of copy operations
+            return f_sequence(copy_results)
+
         # This future completes once *all* copies are done successfully.
         # TODO: this still could be improved, as not every item needs every copy
         # before the state could be refreshed.
         all_copies = f_sequence(copy_results)
 
+        all_copies = f_map(all_copies, fix_failures)
```
rajulkumar, line 323: Single retry might work as expected in current scenario. 
Ideally, the failure that occurred once might occur again. It still the same scenario that could be influenced by number of items that failed, association time pulp is takes and the how frequently the chosen source repo updates.  Hence, this may be retired more than once. The retry count should be defined by some environment variable, that's configured in the target settings.
Other than that this looks good,

rajulkumar (changes requested):

pubtools/_pulp/tasks/push/items/base.py
jm-wk: Thanks, changed.

jm-wk:
Thanks for the comments! Changed the

rajulkumar (dismissed):

rohanpm (changes requested):

pubtools/_pulp/tasks/push/items/base.py
```diff
@@ -281,11 +300,31 @@ def refresh_after_copy(_):
 
             return f
 
+        # A helper function to re-try copying of failed items
+        #  Since the association phase can take a while and the items can be
+        #  moved to other repos meanwhile, we fetch up-to-date info about the
+        #  items from pulp and re-try the failing cases
+        def fix_failures(_):
+            # Get an up-to-date version of all the copy items.
```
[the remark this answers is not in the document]
jm-wk: I have added logging. 

The best way I could think of is returning a tuple with a future of association results and number of failures from fix_failures. I think this is kind of inelegant, is there a better way I could resolve this?

pubtools/_pulp/tasks/push/phase/constants.py
```diff
@@ -112,3 +112,8 @@ def atom(name):
 
 OUT_MAX_FUTURES = int(os.getenv("PUBTOOLS_PULP_OUT_MAX_FUTURES") or "10")
 """Max number of pending futures in output buffer."""
+
+# The following tune the behaviour of individual phases
+
+ASSOCIATION_RETRY_COUNT = int(os.getenv("PUBTOOLS_PULP_ASSOCIATION_RETRY_COUNT") or "1")
```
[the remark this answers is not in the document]
jm-wk: Done.

tests/push/test_association_race.py
```diff
@@ -0,0 +1,102 @@
...
+        # It should be able to run without crashing.
+        command_tester.test(
+            run,
+            args,
+            compare_plaintext=False,
+            compare_jsonl=False,
+            # This will ensure the Pulp state matches the baseline.
+            compare_extra=compare_extra,
+        )
+
```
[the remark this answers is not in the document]
jm-wk: Done.

jm-wk:
Thanks for the comments!

pubtools/_pulp/tasks/push/items/base.py
```diff
@@ -281,11 +300,42 @@ def refresh_after_copy(_):
...
+            return f_sequence(copy_results), failure_count
+
         # This future completes once *all* copies are done successfully.
         # TODO: this still could be improved, as not every item needs every copy
         # before the state could be refreshed.
         all_copies = f_sequence(copy_results)
 
+        # If there are any failed copy operations, retry them
+        for i in range(association_retry_count):
+            if len(all_copies.result()) > 0:
```
rohanpm, line 328: The call to `.result()` here will cause the whole method to become blocking, which is probably going to slow down large pushes a lot.

The caller of this (Associate phase) looks like this:

```
    def run(self):
        for batch in self.iter_for_associate():
            for items in PulpPushItem.items_by_type(batch):
                for associated_f in PulpPushItem.associated_items_single_batch(
                    self.pulp_client, items, self.copy_options
                ):
                    self.put_future_outputs(associated_f)
```

Prior to your change the behavior was:
- associated_items_single_batch starts copy of a batch and returns a future.
- Associate arranges for the outputs to be sent to the next phase when that future completes.

Since this loop only starts for (and doesn't await) completion of batches, it can start processing of more than one batch at once.

Now with your change the behavior would be:

- associated_items_single_batch starts copy of a batch, *waits for it to complete* (because `.result()` was called) and returns a future.
- Associate arranges for the outputs to be sent to the next phase when that future completes (which is always immediately).

That changes the loop so that it's only able to process a single batch (of default size 1,000 items) at a time.  I checked a completed large push at https://url.corp.redhat.com/x and Associate was working on up to 29,000 items at once, so I think if there is a change which causes it to only handle 1,000 items at once there will be a significant performance impact.

I think this needs to be reworked, the retry has to happen *inside* of the future monad rather than awaiting completion of the future.  The way you had it earlier seemed more correct in this regard (but had the problems of not logging & not checking if we actually had to retry).

rohanpm (changes requested):

pubtools/_pulp/tasks/push/items/base.py
jm-wk: Thanks for the explanation! I have completely missed that. 

I have reworked the PR, so that the retry operation is not blocking.

pubtools/_pulp/tasks/push/items/base.py
```diff
@@ -281,11 +300,42 @@ def refresh_after_copy(_):
 
             return f
 
+        # A helper function to re-try copying of failed items
+        #  Since the association phase can take a while and the items can be
+        #  moved to other repos meanwhile, we fetch up-to-date info about the
+        #  items from pulp and re-try the failing cases
+        def retry_failed_items(_):
+            copy_results = []
+            for i in range(association_retry_count):
```
rohanpm, line 309: I don't think the loop is going to work as desired in the case where more than 1 retry happens.  Correct me if I'm wrong but I think it'll work like this:

- first iter (i=0)
- get state of all pulp items
- let's say that enqueue_items returns non-empty copy_results, so some copies are being retried
- we log "Retrying association"
- then *we do not wait for copy to complete*, so we immediately continue on to...
- second iter (i=1)
- get state of all pulp items
- given that we've only just started a new copy and didn't wait for it to complete, the state is probably going to be the same as it was at i=0, so we're going to retry the same copy again.

By my reading this loop is always going to queue retries up to the configured retry count as fast as possible, or have I missed something?

If the loop is aiming to achieve "retry copy until it succeeds" I'd expect somewhere in the loop body it'll have to wait on the results, probably just after the log message?  Something like `f_sequence(copy_results).result()` ?

Note it'd generally be a bad practice for a function like this to block and call `.result()`, but I think it's OK in this case since retrying should be a niche scenario.

tests/push/test_association_race.py
```diff
@@ -0,0 +1,110 @@
...
+    # We can determine that publish didn't occur by checking all
+    # encountered states of push items.
+    all_states = set([item["state"] for item in stub_collector])
+
+    # Everything should be either PENDING (before upload to Pulp)
+    # or EXISTS (after upload), but nothing should be PUSHED since
+    # publish didn't happen.
+    assert all_states == set(["PENDING", "EXISTS"])
+
+    # Assert there was exactly one retry of association.
```
rohanpm, line 105: Comment says you want to assert there was exactly one retry, but then the actual assertions are that there are two retries.

Why are there two retries anyway?  Does the retried association succeed eventually, and if so why's that after 2 retries instead of just 1?

rohanpm (changes requested):

tests/push/test_association_race.py
jm-wk: Here I am asserting that there is one attempt and that the second one is not logged.

Unfortunately it seems that the copy operation is done so quickly that this failed to catch the case of

pubtools/_pulp/tasks/push/items/base.py
jm-wk: Thank you for the comment! I have completely missed that. 

I have added `f_sequence(copy_results).result()` to the loop.

tests/push/test_association_race.py
rohanpm: Sorry, I've misread the code before posting this.

rohanpm (approved):

rajulkumar (approved):

PR #247, opened 2023-10-19
Do garbage collect in batches [RHELDST-20725] 

amcmahon-rh:
Scheduled garbage collection jobs were failing because of out of memory exceptions. This change makes garbage collect run in batches to limit the amount of resources used at once.

### DIFF
--- a/pubtools/_pulp/tasks/garbage_collect.py
+++ b/pubtools/_pulp/tasks/garbage_collect.py
@@ -1,3 +1,4 @@
+import os
 import logging
 from datetime import datetime, timedelta
 
@@ -10,6 +11,8 @@ from pubtools._pulp.services import PulpClientService
 LOG = logging.getLogger("pubtools.pulp")
 step = PulpTask.step
 
+UNASSOCIATE_BATCH_LIMIT = int(
+    os.getenv("PULP_GC_UNASSOCIATE_BATCH_LIMIT", "10000"))
 
 class GarbageCollect(PulpClientService, PulpTask):
     """Perform garbage collection on Pulp data.
@@ -96,16 +99,28 @@ class GarbageCollect(PulpClientService, PulpTask):
             LOG.info("No repos found for cleaning.")
             return
         arc_repo = clean_repos[0]
-        deleted_arc = list(arc_repo.remove_content(criteria=criteria))
+
         deleted_content = []
-        for task in deleted_arc:
-            if task.repo_id == "all-rpm-content":
+
+        while True:
+            deletion_tasks = arc_repo.remove_content(
+                criteria=criteria, limit=UNASSOCIATE_BATCH_LIMIT
+            ).result()
+            arc_tasks = [t for t in deletion_tasks if
+                         t.repo_id == "all-rpm-content"]
+            for task in arc_tasks:
                 for unit in task.units:
                     LOG.info("Old all-rpm-content deleted: %s", unit.name)
                     deleted_content.append(unit)
-        if not deleted_content:
-            LOG.info("No all-rpm-content found older than %s", arc_threshold)
 
+            if not arc_tasks or \
+                    any([t for t in arc_tasks
+                         if len(t.units) < UNASSOCIATE_BATCH_LIMIT]):
+                break
+
+        if not deleted_content:
+            LOG.info("No all-rpm-content found older than %s",
+                     arc_threshold)
 
 def entry_point():
     with GarbageCollect() as instance:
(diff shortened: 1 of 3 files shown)


pubtools/_pulp/tasks/garbage_collect.py
```diff
@@ -96,13 +99,32 @@ def clean_all_rpm_content(self):
...
+            for task in arc_tasks:
                 for unit in task.units:
                     LOG.info("Old all-rpm-content deleted: %s", unit.name)
                     deleted_content.append(unit)
+
+            # Assuming the client can return multiple (python) tasks.
+            # If all tasks delete less units than the batch limit,
+            # implying all relevant units have been unassociated.
+            if not arc_tasks or not any(
+                [t for t in arc_tasks if len(t[0].units) < UNASSOCIATE_BATCH_LIMIT]
```
rohanpm, line 124: I don't get the `t[0]` here.  What is the type of t?  Isn't it supposed to be a `Task`?

In the loop just above, we do `for task in arc_tasks` and then access `task.units`.  I don't get how it could be valid to access this as both `task.units` and `task[0].units`.

pubtools/_pulp/tasks/garbage_collect.py
```diff
@@ -96,13 +99,32 @@ def clean_all_rpm_content(self):
...
+            )
+            for task in arc_tasks:
                 for unit in task.units:
                     LOG.info("Old all-rpm-content deleted: %s", unit.name)
                     deleted_content.append(unit)
+
+            # Assuming the client can return multiple (python) tasks.
+            # If all tasks delete less units than the batch limit,
+            # implying all relevant units have been unassociated.
+            if not arc_tasks or not any(
```
rohanpm, line 123: Not sure about the second half of this condition.

In pseudocode this all seems to mean:

```
if
    not (there were tasks)
    or
    not (there was a task which unassociated less than the limit)
then stop looping
```

Should that second `not` really be there?  If there *was not* a task which unassociated less than the limit, isn't that an indication that we need to keep going as there are likely still units to be processed?

rohanpm (changes requested):

pubtools/_pulp/tasks/garbage_collect.py
```diff
@@ -96,13 +99,29 @@ def clean_all_rpm_content(self):
...
-        for task in deleted_arc:
-            if task.repo_id == "all-rpm-content":
+
+        while True:
+            delete_arc_tasks = list(
+                arc_repo.remove_content(
+                    criteria=criteria, limit=UNASSOCIATE_BATCH_LIMIT
+                )
+            )
+            # The pulp client returns task(s) wrapped in a list
```
rohanpm, line 111: Still not really sure what this comment means.  I feel like this comment and the surrounding code are making things out to be more complicated than they are.

`remove_content` returns a `Future[list[Task]]`, the Future has already been unpacked at this point implicitly by the `list(...)` in the previous statement, so we just have a `list[Task]` here.  To me saying that it's "wrapped" in a list would indicate that there's a redundant level of list.

If I think about another common list for a moment: command-line arguments are passed as a `list[str]` in `sys.argv`.  Would it be natural to say that the arguments are "wrapped" in a list?  To me it's not, they just natively *are* a list.  If someone wrote code dealing with sys.argv with a comment that arguments were wrapped in a list, I'd think they're trying to call attention to something unusual, like the code having to unexpectedly deal with `list[list[str]]` and having to "unwrap" it to get back to the actual argument list of `list[str]`.

Overall, this comment gives an impression of type confusion.  It might be less confusing if the code around here instead looked like:

```python
delete_arc_tasks = arc_repo.remove_content(
    criteria=criteria, limit=UNASSOCIATE_BATCH_LIMIT
).result()
arc_tasks = [t for t in delete_arc_tasks if t.repo_id == "all-rpm-content"]
for task in arc_tasks:
```

rohanpm (approved):

pubtools/_pulp/tasks/garbage_collect.py
amcmahon-rh: A lot of the pulp client functions return a single Task inside a list, which to me felt odd. Though I guess it's by design, in case some functions were intended to return multiple Tasks.

rohanpm (approved):

pubtools/_pulp/tasks/garbage_collect.py
rohanpm: Right, the reason for that is to avoid locking the methods into generating precisely one task, as this way they can be extended without breaking backwards-compatibility.

For example, for this particular case it would also have been possible to update `remove_content` to internally do the pagination, rather than requiring the caller to do it.  That would have created one task per "page", and since the return value has always been a list of tasks, it'd be backwards-compatible.  Maybe it's a shame we didn't do it that way actually, I didn't think of that at the beginning.

The same applies to other things like:

- `copy_content` possibly ought to do pagination internally as well
- various methods might want to have side-effects added along the lines of: every time you do X to some repo, it should also automatically be published.

PR #257, opened 2023-11-21
Do batching with unit IDs [RHELDST-20725]

amcmahon-rh:
We want to do garbage collection in batches since doing it all at once can cause OOM exceptions. The original attempt used the limit value in criteria submitted to the unassociation endpoint, which did not work. This change instead creates batches using unit searching instead. The size of the batches can be changed by setting the
PUBTOOLS_PULPLIB_PAGE_SIZE env-var.

### DIFF
--- a/pubtools/_pulp/tasks/garbage_collect.py
+++ b/pubtools/_pulp/tasks/garbage_collect.py
@@ -11,8 +11,7 @@ from pubtools._pulp.services import PulpClientService
 LOG = logging.getLogger("pubtools.pulp")
 step = PulpTask.step
 
-UNASSOCIATE_BATCH_LIMIT = int(
-    os.getenv("PULP_GC_UNASSOCIATE_BATCH_LIMIT", "10000"))
+UNASSOCIATE_BATCH_LIMIT = int(os.getenv("PULP_GC_UNASSOCIATE_BATCH_LIMIT", "10000"))
 
 
 class GarbageCollect(PulpClientService, PulpTask):
@@ -98,8 +97,7 @@ class GarbageCollect(PulpClientService, PulpTask):
             Criteria.with_unit_type(RpmUnit, unit_fields=["unit_id"]),
             Criteria.with_field(
                 "cdn_published",
-                Matcher.less_than(
-                    datetime.utcnow() - timedelta(days=arc_threshold)),
+                Matcher.less_than(datetime.utcnow() - timedelta(days=arc_threshold)),
             ),
         )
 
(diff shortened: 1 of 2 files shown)


pubtools/_pulp/tasks/garbage_collect.py
```diff
@@ -96,13 +90,45 @@ def clean_all_rpm_content(self):
...
-                    LOG.info("Old all-rpm-content deleted: %s", unit.name)
-                    deleted_content.append(unit)
+
+        LOG.info("Collecting unit batches for deletion")
+        first_page = arc_repo.search_content(criteria=search_criteria)
+
+        def handle_pages(page):
+            batch_accumulator = [[unit.unit_id for unit in page.data]]
+            if page.next:
+                batch_accumulator.extend(f_flat_map(page.next, handle_pages).result())
```
rohanpm, line 109: We'll need at least one test which generates more than one page of data so this will be covered.

Incidentally, I'm not sure there is much point of writing the code this way with flat_map because you immediately call 'result()` on the handler for the next page, making the whole thing blocking anyway, and making the call stack N deep if there are N pages.  So this is kind of paying the cost of having more complicated code dealing with futures, without getting the benefit as every future is immediately awaited and nothing is allowed to run concurrently.

This simpler approach might have worked (haven't tested it) and should allow some things to run concurrently:

```python
# Wait for the first page
page = arc_repo.search_content(criteria=search_criteria).result()

# While there is a page of data to be processed...
while page:
  # process it
  batch = page.data
  do_something_with(batch)

  # Wait for the next page (if there is one).
  # note: the client has been working on fetching the next page in the background
  # even before we get here.
  page = page.next.result() if page.next else None
```

Or, another alternative:

```python
batch = []
batch_size = 1000
for unit in arc_repo.search_content(criteria=search_criteria):
  batch.append(unit)
  if len(batch) > batch_size:
    process_batch(batch)
    batch = []
if batch:
  process_batch(batch)
```

rohanpm (changes requested):

pubtools/_pulp/tasks/garbage_collect.py
amcmahon-rh: I'd like to process the search results into batches and then do the deletion afterwards. The search_content pagination seems to make requests with the limit/ skip values, so mixing  in deletion requests with that feels dangerous. I've simplified code to arrange the batches and added in a test.

tests/garbage_collect/test_garbage_collect.py
```diff
@@ -227,11 +230,74 @@ def test_arc_garbage_collect(mock_logger):
...
             gc.main()
+
     updated_rpm = list(client.get_repository("all-rpm-content").search_content())
     assert len(updated_rpm) == 1
     mock_logger.info.assert_any_call("Old all-rpm-content deleted: %s", rpm2.name)
 
 
+def test_arc_garbage_collect_in_batches(mock_logger):
+    """deletes relevant all-rpm-content content in batches"""
+    gc_module.UNASSOCIATE_BATCH_LIMIT = 5
```
rohanpm, line 241: Nothing will automatically undo this at the end of the test, so if any other tests are influenced by UNASSOCIATE_BATCH_LIMIT, their behavior will differ depending on whether the current process has run this test first.  Even if it doesn't affect any existing test, it's a trap for developers adding more tests later.

It's best to avoid that and have tests undo whatever changes they make.  You can use monkeypatch fixture to do this easily, something like:

```python
def test_arc_garbage_collect_in_batches(mock_logger, monkeypatch: pytest.MonkeyPatch):
    """deletes relevant all-rpm-content content in batches"""
    monkeypatch.setattr(gc_module, 'UNASSOCIATE_BATCH_LIMIT', 5)
```

Then pytest will take care of undoing this for you at the end of the test.

rohanpm (changes requested):

rohanpm (approved):

PR #261, opened 2024-02-02
Retry copy operation [RHELDST-18718] 

rbikar:
Added retry mechanism for items that were not copied
successfully on the first attempt during pubtools-push.

The new approach roughly performs this extra operations:
* wait for all copies to finish
* refresh items states to current stage in pulp
* yield items that were copied successfully
* retry whole copy operation with unsuccessfully copied items
* retry count is set to 5 as default

I also did a little refactor in `items_with_pulp_state_single_batch` -
I moved 'preparation of items' and 'copies submit' to their own
private methods because the original method grew a bit.


rbikar:
This PR should fix the first attempt of adding retries in https://github.com/release-engineering/pubtools-pulp/pull/226 where we actually hit deadlock. I wasn't able to pinpoint the exact place of deadlock but the code seemed to be too nested and chained with futures. So hopefully this new approach will not hit any deadlock as the code is more straightforward. 
The race-condition test was copied from the original https://github.com/release-engineering/pubtools-pulp/pull/226

Note on failed tests - I'm not sure what's exactly wrong there, locally tests on py312 seem to work just fine. It looks like there some issue with yaml.
```
raise RepresenterError("cannot represent an object", data)
E       yaml.representer.RepresenterError: ('cannot represent an object', frozendict.frozendict({'default': frozendict.frozendict({'description': 'The minimal, distribution-compiled dwm binary.', 'rpms': ['dwm']}), 'user': frozendict.frozendict({'description': 'Includes distribution-compiled dwm as well as a helper script to apply user patches and configuration, dwm-user.', 'rpms': ['dwm', 'dwm-user']})}))
``` but I haven't dug into it.

rohanpm (approved):

PR #262, opened 2024-02-07
Implement copy_repo entry point [RHELDST-549]

negillett:
Given a list of comma-separated pairs of repos, copy-repo will copy content from the first to the second in the pair. Supports --skip and --content-type.

`copy-repo --target <x> src_a,dst_a src_b,dst_b`

### DIFF
--- a/pubtools/_pulp/tasks/common.py
+++ b/pubtools/_pulp/tasks/common.py
@@ -6,13 +6,20 @@ import sys
 
 import attr
 from more_executors.futures import f_flat_map, f_map, f_return, f_sequence
-from pubtools.pulplib import (ErratumUnit, FileUnit, ModulemdUnit,
-                              PublishOptions, RpmUnit)
-from pushsource import (ErratumPushItem, FilePushItem, ModuleMdPushItem,
-                        RpmPushItem)
+from pubtools.pulplib import (
+    ErratumUnit,
+    FileUnit,
+    ModulemdUnit,
+    PublishOptions,
+    RpmUnit,
+)
+from pushsource import ErratumPushItem, FilePushItem, ModuleMdPushItem, RpmPushItem
 
-from pubtools._pulp.services import (CdnClientService, FastPurgeClientService,
-                                     UdCacheClientService)
+from pubtools._pulp.services import (
+    CdnClientService,
+    FastPurgeClientService,
+    UdCacheClientService,
+)
 from pubtools._pulp.task import PulpTask
 
 from ..hooks import pm
--- a/pubtools/_pulp/tasks/common.py
+++ b/pubtools/_pulp/tasks/common.py
@@ -1,18 +1,19 @@
-import os
-import logging
+import collections
 import datetime
+import logging
+import os
+import sys
+
 import attr
+from more_executors.futures import f_flat_map, f_map, f_return, f_sequence
+from pubtools.pulplib import (ErratumUnit, FileUnit, ModulemdUnit,
+                              PublishOptions, RpmUnit)
+from pushsource import (ErratumPushItem, FilePushItem, ModuleMdPushItem,
+                        RpmPushItem)
 
-from more_executors.futures import f_map, f_flat_map, f_sequence, f_return
-
-from pubtools.pulplib import PublishOptions
-
+from pubtools._pulp.services import (CdnClientService, FastPurgeClientService,
+                                     UdCacheClientService)
 from pubtools._pulp.task import PulpTask
-from pubtools._pulp.services import (
-    FastPurgeClientService,
-    UdCacheClientService,
-    CdnClientService,
-)
 
 from ..hooks import pm
 
@@ -82,13 +83,15 @@ class UdCache(UdCacheClientService):
     @step("Flush UD cache")
     def flush_ud(self, repos, errata=None):
         client = self.udcache_client
+        out = []
         if not client:
             LOG.info("UD cache flush is not enabled.")
-            return []
+            return out
 
-        out = []
         for repo in repos:
             out.append(client.flush_repo(repo.id))
+            if repo.eng_product_id:
+                out.append(client.flush_product(repo.eng_product_id))
 
         out.extend([client.flush_erratum(erratum.id) for erratum in (errata or [])])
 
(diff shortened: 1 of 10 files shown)


negillett:
Hmm. I don't see doc_parser (the missing coverage) being tested for other tasks. Doesn't seem beneficial but I can add a test anyway.

negillett:
> Hmm. I don't see doc_parser (the missing coverage) being tested for other tasks. Doesn't seem beneficial but I can add a test anyway.

Nevermind. I found shared task tests.

negillett:
...just found tasks/push/copy. Did I waste my time writing this (adapting it from clear_repo)?

pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -0,0 +1,324 @@
...
+            help="repository pair(s) (src, dest) to be copied. e.g. repo-A,repo-B repo-C,repo-D",
+        )
+
+    def fail(self, *args, **kwargs):
+        LOG.error(*args, **kwargs)
+        sys.exit(30)
+
+    @step("Check repos")
+    def get_repos(self, repos):
+        # Returns all pairs of repos to be operated on by this task.
```
rohanpm, line 81: This comment doesn't seem accurate.  It looks like `repos` is expected to be a single pair only.  If this were not the case, the code a bit later on which looks like `if repo.id == repos[0]:` would be wrong as there is never any attempt to process anything beyond the first two elements of `repos`.

If it's designed to work on only a single pair at a time, may I suggest that `get_repos(self, src_repo, dst_repo):` or similar would be clearer?

pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -0,0 +1,324 @@
...
+
+        return copied_repo
+
+    @step("Copy content")
+    def copy_content(self, src_repo, dest_repo):
+        futures = []
+        content_types = self.content_type or ["None"]
+        for t in content_types:
+            crit = None
+            if t == "iso":
```
rohanpm, line 159: It feels like "erratum" might be missing here.  Generally if we're copying RPMs, we'd want to be copying the erratum they belong to as well, so it seems quite suspicious to find code which supports ("rpm", "srpm") but not "erratum".

pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -0,0 +1,324 @@
...
+        for unit in task.units:
+            push_item = self.push_item_for_unit(unit)
+            if push_item:
+                out.append(push_item)
+        return out
+
+    def push_item_for_unit(self, unit):
+        for unit_type, fn in [
+            (ModulemdUnit, self.push_item_for_modulemd),
+            (RpmUnit, self.push_item_for_rpm),
```
rohanpm, line 194: I think this is another place where ErratumUnit possibly ought to be supported?

pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -0,0 +1,324 @@
...
+            (FileUnit, self.push_item_for_file),
+        ]:
+            if isinstance(unit, unit_type):
+                return fn(unit)
+
+    def push_item_for_modulemd(self, unit):
```
rohanpm, line 200: If these methods have been copy-pasted, can you please find a way to share them instead of duplicating them?

pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -0,0 +1,324 @@
...
+        out["origin"] = "pulp"
+        out["name"] = unit.path
+        out["sha256sum"] = unit.sha256sum
+
+        return FilePushItem(**out)
+
+    @step("Publish")
+    def publish(self, repo_fs):
+        return [
+            f_flat_map(f, lambda r: r.publish(PublishOptions(clean=True)))
```
rohanpm, line 258: I think you can remove clean=True, that would be specific to the "delete" use-case, I can't think of any reason it's needed after a copy.

pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def run(self):
+        to_await = []
+
+        # Start copying repos.
+        repos_to_flush = []
+        repo_copies_fs = []
+        for pair in self.repo_pairs:
+            LOG.info("Copying repo %s to %s", pair[0], pair[1])
+            src_repo, dest_repo = self.get_repos(pair)
```
rohanpm, line 285: Relating to my earlier comment about what `get_repos` is supposed to accept:

The code here is why I'm saying it looks like it's supposed to accept just one pair, and not a list of pairs like a comment within the method suggests.

However, I think you probably should change it so it does accept a list of pairs.  I think the main point of that code was to verify that the user had provided valid IDs for all of their repos before we start doing any real work.  The way this loop is set up, it won't accomplish that now. It could be fixed if get_repos indeed accepted a list of pairs of repo IDs, `list[tuple[str,str]]` and returned a list of pairs of repos `list[tuple[Repository,Repository]]`.

For example if the user passes "src1,dst1 src2,dst2 src3,dst3" and src3 doesn't exist, we ideally want the task to quickly determine that and exit with an error.  Right now it looks like it would start the copies between src1/dst1 and src2/dst2 before crashing on src3.

tests/copy_repo/test_copy_repo.py
```diff
@@ -0,0 +1,387 @@
...
+            ],
+        )
+
+    # It should record that it copied these push items:
+    assert sorted(fake_collector.items, key=lambda pi: pi["filename"]) == [
+        {
+            "state": "PUSHED",
+            "origin": "pulp",
+            "src": None,
+            "dest": None,
```
rohanpm, line 146: Why are the src and dest 'None' ?

Well, src I'm not entirely sure it's appropriate for it to be non-None.  But at least shouldn't `dest` be the destination repo of the copy?

rohanpm (changes requested):

rohanpm:
> ...just found tasks/push/copy. Did I waste my time writing this (adapting it from clear_repo)? @rajulkumar, @rbikar, @rohanpm

No, that file only has some utility functions used during copy steps in other tasks.  It's far from a complete implementation of a command dedicated to copying.

pubtools/_pulp/tasks/copy_repo.py
negillett: Thanks!

tests/copy_repo/test_copy_repo.py
```diff
@@ -0,0 +1,411 @@
...
+    """Command fails when pointed at nonexistent repos."""
+    # This one test covers the entry point
+
+    command_tester.test(
+        lambda: pubtools._pulp.tasks.copy_repo.entry_point(FakeCopyRepo),
+        [
+            "test-copy-repo",
+            "--pulp-url",
+            "https://pulp.example.com/",
+            "repo1,repo2 repo3,repo4",
```
rohanpm, line 70: This part looks a bit suspicious.

When a person invokes this command, the repo pairs are likely to come in as separate arguments, e.g. the list would look like:

```
            "test-copy-repo",
            "--pulp-url",
            "https://pulp.example.com/",
            "repo1,repo2",
            "repo3,repo4",
```

It only wouldn't be that way if, for example, the user ran it from a shell and explicitly quoted it like:

```
pubtools-pulp-copy-repo --pulp-url https://pulp.example.com/ "repo1,repo2 repo3,repo4"
```

...and I wouldn't expect them to do that (and they shouldn't have to).

If the code works correctly in both cases I think that's fine, but I'd say the test should focus on the more common case, which should be the case of repo1,repo2 and repo3,repo4 arriving as two separate arguments.

pubtools/_pulp/tasks/common.py
```diff
@@ -182,3 +190,161 @@ def publish_with_cache_flush(
...
+
+        for task in repo.tasks:
+            for unit in task.units:
+                type_id = unit.content_type_id
+                content_types[type_id] = content_types.get(type_id, 0) + 1
+
+        task_ids = ", ".join(sorted([t.id for t in repo.tasks]))
+        repo_id = repo.repo.id
+        if not content_types:
+            LOG.warning("%s: no content removed, tasks: %s", repo_id, task_ids)
```
rohanpm, line 231: This looks slightly wrong as you don't know if the action was 'remove'. I guess it's going to give this warning even for copies, which will confuse people.

pubtools/_pulp/tasks/common.py
```diff
@@ -182,3 +190,161 @@ def publish_with_cache_flush(
...
+        else:
+            removed_types = []
```
rohanpm, line 233: Similar to the above comment, since we don't know if this was actually a remove operation, some more general term like "affected_types" may be more appropriate and less confusing to the reader.

rohanpm (commented):
Remember to update setup.py to add the new entry point under console_scripts.

tests/copy_repo/test_copy_repo.py
rohanpm: I took a look at the implementation and I don't think it actually does support this way, so I think this test only works because it's the test for missing repos.  It seems like internally it'd be splitting this into "repo1", "repo2 repo3", "repo4".  The test may 'work' since "repo2 repo3" isn't an existing repo but it wouldn't be what you intended.

Speaking of which, while reading that code I also noted that it doesn't really detect if the user messes up the format of the argument, so I'll comment about that too.

pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -0,0 +1,200 @@
...
+        # Otherwise, return None to let library defaults apply
+        c = self.args.content_type or None
+        # Normalize content types (e.g., "ISO" -> "iso").
+        if c:
+            c = [t.lower() for t in c]
+        return c
+
+    @property
+    def repo_pairs(self):
+        return [pair.split(",") for pair in self.args.repopairs]
```
rohanpm, line 49: I think either here, or in `get_repos`, we ought to be validating that we really have pairs.

I'm reading through how this is used in `get_repos` and it seems pretty easy for the caller to mess up and have the command not do what was intended.  For example if the user accidentally types:

```
copy-repo ... repo1,repo2,repo3,repo4
```

...where they meant to type `repo1,repo2 repo3,repo4`, it won't detect a problem.  It seems like it'll just do the copy between repo1,repo2 and ignore repo3,repo4 with no indication to the user that something went wrong.

tests/copy_repo/test_copy_repo.py
negillett: Ah, okay. Yeah, I wrote it for that case because that's what was presented in the Jira description. I'll make sure it's capable of both.

pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def log_copy(self, copied_repo):
+        # Given a repo which has been copied, log some messages
+        # summarizing the copied unit(s)
+        content_types = {}
+
+        for task in copied_repo.tasks:
+            for unit in task.units:
+                type_id = unit.content_type_id
+                content_types[type_id] = content_types.get(type_id, 0) + 1
```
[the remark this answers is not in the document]
negillett: This bit was pasted from clear_repo but I can make the improvement.

pubtools/_pulp/tasks/common.py
negillett: Thanks!

setup.py
```diff
@@ -47,6 +47,7 @@ def get_requirements():
             "pubtools-pulp-push = pubtools._pulp.tasks.push:entry_point",
             "pubtools-pulp-fix-cves = pubtools._pulp.tasks.fix_cves:entry_point",
             "pubtools-pulp-delete = pubtools._pulp.tasks.delete:entry_point",
+            "pubtools-pulp-copy-repo = pubtools._pulp.tasks.delete:entry_point",
```
rohanpm, line 50:
```suggestion
            "pubtools-pulp-copy-repo = pubtools._pulp.tasks.copy_repo:entry_point",
```

rohanpm (approved):

PR #265, opened 2024-02-20
copy-repo: accept modulemd_default as --content-type

negillett:


pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -137,15 +137,15 @@ def get_repos(self):
...
+            if content_type == "iso":
                 crit = Criteria.with_unit_type(FileUnit)
-            if t in ("rpm", "srpm"):
+            if content_type in ("rpm", "srpm"):
                 crit = Criteria.with_unit_type(RpmUnit)
-            if t == "erratum":
+            if content_type == "erratum":
                 crit = Criteria.with_unit_type(ErratumUnit)
-            if t == "modulemd":
+            if content_type in ("modulemd", "modulemd_default"):
```
rohanpm, line 148: I think there are a couple of issues with this:

First, the minor, it seems the option should be "modulemd_defaults" (with an "s"), not "modulemd_default".  The actual unit type ID as used by Pulp has the trailing "s".

Second, is it really correct to just process ModulemdUnit for both these values?  There is a separate class and unit type for "modulemd_defaults".  The pubtools-pulplib class is ModulemdDefaultsUnit.  I would have expected this to require a separate branch for each content_type and unit class, same as most of the other branches work.

rohanpm (changes requested):

PR #266, opened 2024-03-05
Update publish_with_cache_flush in common.py

lekang9:
- When publish is skipped, we should also skip cdn cache flush
- Update clear_repo.py to use publish_with_cache_flush method instead

### DIFF
--- a/pubtools/_pulp/step.py
+++ b/pubtools/_pulp/step.py
@@ -9,8 +9,10 @@ LOG = logging.getLogger("pubtools.pulp")
 class StepDecorator(object):
     """Implementation of PulpTask.step decorator. See that method for more info."""
 
-    def __init__(self, name):
+    def __init__(self, name, depends_on, skipped_value):
         self._name = name
+        self._depends_on = depends_on or []
+        self._skipped_value = skipped_value
 
     @property
     def human_name(self):
@@ -28,7 +30,11 @@ class StepDecorator(object):
                     self.human_name,
                     extra={"event": {"type": "%s-skip" % self.machine_name}},
                 )
-                return args[0] if args else None
+                return (
+                    self.skipped_value
+                    if self._skipped_value is not UNSET
+                    else args[0] if args else None
+                )
 
             logger = StepLogger(self)
             args = logger.log_start(args)
@@ -53,7 +59,8 @@ class StepDecorator(object):
 
     def should_skip(self, instance):
         skip = (getattr(instance.args, "skip", None) or "").split(",")
-        return self.machine_name in skip
+        deps_and_self = self._depends_on + [self.machine_name]
+        return any([name in skip for name in deps_and_self])
 
 
 # helpers used in implementation of decorator
--- a/pubtools/_pulp/step.py
+++ b/pubtools/_pulp/step.py
@@ -6,6 +6,7 @@ from more_executors.futures import f_sequence, f_return
 LOG = logging.getLogger("pubtools.pulp")
 UNSET = object()
 
+
 class StepDecorator(object):
     """Implementation of PulpTask.step decorator. See that method for more info."""
 
--- a/pubtools/_pulp/step.py
+++ b/pubtools/_pulp/step.py
@@ -4,7 +4,7 @@ import inspect
 from more_executors.futures import f_sequence, f_return
 
 LOG = logging.getLogger("pubtools.pulp")
-
+UNSET = object()
 
 class StepDecorator(object):
     """Implementation of PulpTask.step decorator. See that method for more info."""
@@ -31,7 +31,7 @@ class StepDecorator(object):
                     extra={"event": {"type": "%s-skip" % self.machine_name}},
                 )
                 return (
-                    self.skipped_value
+                    self._skipped_value
                     if self._skipped_value is not UNSET
                     else args[0] if args else None
                 )
--- a/pubtools/_pulp/tasks/common.py
+++ b/pubtools/_pulp/tasks/common.py
@@ -36,7 +36,7 @@ step = PulpTask.step
 class CDNCache(FastPurgeClientService, CdnClientService):
     """Provide features to interact with CDN cache."""
 
-    @step("Flush CDN cache")
+    @step("Flush CDN cache", depends_on=["publish"], skipped_value=[])
     def flush_cdn(self, repos):
         """Clears the CDN cache for the repositories provided
 
--- a/pubtools/_pulp/task.py
+++ b/pubtools/_pulp/task.py
@@ -4,7 +4,7 @@ from argparse import ArgumentParser, RawDescriptionHelpFormatter
 
 from pubtools.pluggy import task_context
 
-from .step import StepDecorator
+from .step import StepDecorator, UNSET
 
 
 LOG = logging.getLogger("pubtools.pulp")
@@ -68,13 +68,17 @@ class PulpTask(object):
         return self._args
 
     @classmethod
-    def step(cls, name):
+    def step(cls, name, depends_on=None, skipped_value=UNSET):
         """A decorator to mark an instance method as a discrete workflow step.
 
         Marking a method as a step has effects:
 
         - Log messages will be produced when entering and leaving the method
         - The method can be skipped if requested by the caller (via --skip argument)
+        - Methods that depend on other methods are implicitly skipped if depends_on
+          is supplied and that dependant is being skipped
+        - If the method is skipped, it returns either `skipped_value` (if that has
+          been set), or the method's first argument.
 
         Steps may be written as plain blocking functions, as non-blocking
         functions which accept or return Futures, or as generators.
@@ -94,7 +98,7 @@ class PulpTask(object):
         - The step is considered *failed* if it raised an exception.
         - The step is considered *finished* once all items have been yielded.
         """
-        return StepDecorator(name)
+        return StepDecorator(name, depends_on, skipped_value)
 
     def _basic_args(self):
         # minimum args required for a pulp CLI task
--- a/tests/logs/clear_repo/test_clear_repo/test_clear_file_skip_publish.jsonl
+++ b/tests/logs/clear_repo/test_clear_repo/test_clear_file_skip_publish.jsonl
@@ -7,5 +7,4 @@
 {"event": {"type": "publish-skip"}}
 {"event": {"type": "flush-ud-cache-start"}}
 {"event": {"type": "flush-ud-cache-end"}}
-{"event": {"type": "flush-cdn-cache-start"}}
-{"event": {"type": "flush-cdn-cache-end"}}
+{"event": {"type": "flush-cdn-cache-skip"}}
--- a/tests/logs/clear_repo/test_clear_repo/test_clear_file_skip_publish.txt
+++ b/tests/logs/clear_repo/test_clear_repo/test_clear_file_skip_publish.txt
@@ -9,6 +9,4 @@
 [    INFO] Flush UD cache: started
 [    INFO] UD cache flush is not enabled.
 [    INFO] Flush UD cache: finished
-[    INFO] Flush CDN cache: started
-[    INFO] CDN cache flush is not enabled.
-[    INFO] Flush CDN cache: finished
+[    INFO] Flush CDN cache: skipped


lekang9:
looks like there a few things related to tox and static that need to be looked into so that there are no failing checks

lekang9:
@rohanpm Hi, i'm trying to setup virtualenv to run tox for this pull request to attempt things as i run the tests however i'm having issues building wheel for rhsm do you have any idea or advice?
I attempt to run "pip install ." from root dir of pubtools-pulp
```
      gcc -Wno-unused-result -Wsign-compare -DDYNAMIC_ANNOTATIONS_ENABLED=1 -DNDEBUG -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -fPIC -I/home/lekang/tox/include -I/usr/include/python3.8 -c src/certificate.c -o build/temp.linux-x86_64-3.8/src/certificate.o
      src/certificate.c:46:10: fatal error: openssl/asn1.h: No such file or directory
         46 | #include <openssl/asn1.h>
            |          ^~~~~~~~~~~~~~~~
      compilation terminated.
      error: command 'gcc' failed with exit status 1
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for rhsm
  Running setup.py clean for rhsm
Successfully built pubtools-pulp
Failed to build rhsm
ERROR: Could not build wheels for rhsm, which is required to install pyproject.toml-based projects
```
i created virtualenv in following way
```
[lekang@fedora pubtools-pulp]$ python3.8 -m venv ~/tox  
[lekang@fedora pubtools-pulp]$ source ~/tox/bin/activate
(tox) [lekang@fedora pubtools-pulp]$ python -m pip install --upgrade pip
(tox) [lekang@fedora pubtools-pulp]$ pip install wheel
Collecting wheel
  Using cached wheel-0.42.0-py3-none-any.whl.metadata (2.2 kB)
Using cached wheel-0.42.0-py3-none-any.whl (65 kB)
Installing collected packages: wheel
Successfully installed wheel-0.42.0
...

**Comment by rohanpm** [comment]
@lekang9 

The general rule of thumb is, if you try to "pip install" something and you get a compilation error referring to a missing header file like:

```
fatal error: openssl/asn1.h
```

...then you need to install the corresponding "devel" package on your system in order to get those headers.  Installing using the OS packages, not pip.  For example in this case you probably want to install `openssl-devel`.

**Comment by lekang9** [comment]
@rohanpm that did the trick, thank you i'm able to run tox locally and make adjustments without using the pipeline here

File: pubtools/_pulp/tasks/common.py [hunk 1]
```diff
@@ -179,8 +182,9 @@ def publish_with_cache_flush(
         # hook implementation(s) may now flush pulp-derived caches and datastores
         pm.hook.task_pulp_flush()
 
-        # flush CDN cache
-        out = self.flush_cdn(repos)
+        # flush cdn cache if publish is not being skipped
+        if "publish" not in self.args.skip:
```
lekang9, line 186: @rohanpm i'm trying to figure out how the skip argument might be accessed from within the Publisher class in common.py but it seems to be proving difficult looking for some advice. Currently tests are failing like so for publish/:
```
    def publish_with_cache_flush(
        self, repos, units=None, pulp_client=None, errata=None
    ):
        # Ensure all repos in 'repos' are fully published, and CDN/UD caches are flushed.
        #
        # If 'units' are provided, ensures those units have cdn_published field set after
        # the publish and before the UD cache flush.
        #
        # Initialize out variable as an empty list in the case where publish and cdn
        # cache flush are skipped
        units = units or []
        pulp_client = pulp_client or self.pulp_client
        out = []
    
        # publish the repos found
        publish_fs = self.publish(repos)
    
        # wait for the publish to complete before
        # flushing caches.
        f_sequence(publish_fs).result()
    
        # hook implementation(s) may now flush pulp-derived caches and datastores
        pm.hook.task_pulp_flush()
    
        # flush cdn cache if publish is not being skipped
>       if "publish" not in self.args.skip:
E       AttributeError: 'Namespace' object has no attribute 'skip'

pubtools/_pulp/tasks/common.py:186: AttributeError
```
i have been trying to replicate what is seen here at https://github.com/release-engineering/pubtools-pulp/blob/master/pubtools/_pulp/tasks/push/command.py#L51 to be used in a similar fashion in this Publisher class

any advice is appreciated,
Thanks

rohanpm: @lekang9 

That failure seems to be in test_fix_cves.  The problem is that the argument parser used by the fix-cves command doesn't add a `--skip` argument.  So you can't read it.

Your choices would seem to be:

- Make the fix-cves command also accept `--skip` (I like this option, I would like it to be standard for all commands).  This would be done by adding it to `FixCves.add_args` method.
- Or, change the code here reading `skip` to accept when the attribute is missing, e.g. by using `getattr` rather than accessing it directly.

lekang9: Thanks i ended up going with moving it into PulpTask._basic_args which helped to pass the publish tests, however it looks like the changes i made to clearrepo are still causing failures i'll keep looking into that

lekang9: @rohanpm i have another question unrelated to above, so i've cleared majority of the tests however with the changes to clear_repo.py i am still experiencing some failures due to TypeError where f_flat_map expects a future but this is not an issue where it is used elsewhere, my first suspicion is that the tests for clear_repo need to be adjusted to be more similar to test_publish. I'll push a commit shortly with what i have so far so that the tests update here

rohanpm: @lekang9 , evidently `publish` (and by extension `publish_with_cache_flush`) are designed to accept a `list[Future[Repository]]` and you've changed the code here to now pass it a `list[Repository]` instead.  It might help if the code had type hints, but it wasn't an option when the code was originally written.

Note if you're not sure about the type of things, you can run py.test with `--pdb` argument and it'll drop you into a debug session immediately when a test fails, and you can then inspect various objects.

In this case, in order to call `publish_with_cache_flush` with a `list[Future[Repository]]` as it's designed for, you could write:

```
         # They should be published.
-        publish_fs = self.publish_with_cache_flush(repos)
+        publish_fs = self.publish_with_cache_flush([f_map(f, lambda cr: cr.repo) for f in cleared_repos_fs])
```

lekang9: Thanks that is helpful for working through these tests, i've hit another snag in the tests but this is more of a question specific to clear_repo and whether or not we should be having step "set_cdn_published" included because in the diff seen previously we are not doing so (even when flushing ud cache) imo it seems to make sense. Atleast from seeing the comments in the code it makes sense to have this step before flushing ud cache in general.
here is a diff after i've applied the f_map to publish_with_cache_flush in clear_repo.py
```
>           raise AssertionError(message)
E           AssertionError: Output differs from expected (set UPDATE_BASELINES=1 if intended):
E           --- <output of test>
E           +++ /home/lekang/Repos/pubtools-pulp/tests/logs/clear_repo/test_clear_repo/test_clear_repo_multiple_content_types.txt
E           @@ -7,10 +7,12 @@
E            [    INFO] Record push items: finished
E            [    INFO] Publish: started
E            [    INFO] Publish: finished
E           +[    INFO] Flush CDN cache: started
E           +[    INFO] CDN cache flush is not enabled.
E           +[    INFO] Flush CDN cache: finished
E           +[    INFO] Set cdn_published: started
E           +[    INFO] Set cdn_published: finished
E            [    INFO] Flush UD cache: started
E            [    INFO] UD cache flush is not enabled.
E            [    INFO] Flush UD cache: finished
E           -[    INFO] Flush CDN cache: started
E           -[    INFO] CDN cache flush is not enabled.
E           -[    INFO] Flush CDN cache: finished
```

lekang9: FWIW i've updated baselines but there are 2 tests still failing it looks like there is an issue with the url to be flushed not making it through somewhere e.g.
```
        # It should have flushed these URLs
>       assert sorted(task_instance.fastpurge_client.purged_urls) == [
            "https://cdn.example.com/some/publish/url/mutable1",
            "https://cdn.example.com/some/publish/url/mutable2",
        ]
E       AssertionError: assert [] == ['https://cdn...url/mutable2']
E         
E         Right contains 2 more items, first extra item: 'https://cdn.example.com/some/publish/url/mutable1'
E         
E         Full diff:
E         + []
E         - [
E         -     'https://cdn.example.com/some/publish/url/mutable1',
E         -     'https://cdn.example.com/some/publish/url/mutable2',
E         - ]

tests/clear_repo/test_clear_repo.py:197: AssertionError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/lekang/Repos/pubtools-pulp/tests/clear_repo/test_clear_repo.py(197)test_clear_file_repo()
-> assert sorted(task_instance.fastpurge_client.purged_urls) == [
(Pdb) print(task_instance.fastpurge_client.purged_urls
*** SyntaxError: unexpected EOF while parsing
(Pdb) print(task_instance.fastpurge_client.purged_urls)
[]
```
i can only assume that with my input the relative url to be flushed is being lost somewhere along the lines

rohanpm: > Thanks that is helpful for working through these tests, i've hit another snag in the tests but this is more of a question specific to clear_repo and whether or not we should be having step "set_cdn_published" included because in the diff seen previously we are not doing so (even when flushing ud cache) imo it seems to make sense. 

`set_cdn_published` needs to be called with some units, so I'm not sure what units you could pass in the clear_repo case.  The ones just removed from the repo wouldn't make sense as they're not being published to the CDN.

I don't think it can be *meaningfully* invoked in the clear repo case, but if the method is just getting called, logging those lines but not actually doing anything, I think that's fine.

rohanpm: > FWIW i've updated baselines but there are 2 tests still failing it looks like there is an issue with the url to be flushed not making it through somewhere e.g.

I can help you with this if you push the updated code, but I'm only seeing code from 4 days ago.

lekang9: Sorry i forgot to update it, i've just pushed it up

rohanpm: @lekang9 

Did you notice that the baseline update for test_clear_file_repo.txt included an error?

```
--- a/tests/logs/clear_repo/test_clear_repo/test_clear_file_repo.txt
+++ b/tests/logs/clear_repo/test_clear_repo/test_clear_file_repo.txt
@@ -7,10 +7,6 @@
 [    INFO] Record push items: finished
 [    INFO] Publish: started
 [    INFO] Publish: finished
-[    INFO] Flush UD cache: started
-[    INFO] Flush UD cache: finished
 [    INFO] Flush CDN cache: started
-[    INFO] Flushing cache for some-filerepo:
-[    INFO]    https://cdn.example.com/some/publish/url/mutable1
-[    INFO]    https://cdn.example.com/some/publish/url/mutable2
-[    INFO] Flush CDN cache: finished
+[   ERROR] Flush CDN cache: failed
+# Raised: 'MapFuture' object has no attribute 'relative_url'
```

That exception is what needs to be resolved first, don't worry about the assertion.

The error can also be seen if you run the test with `--log-cli-level=DEBUG`

```
$ py.test -xv tests/clear_repo/test_clear_repo.py::test_clear_file_repo --log-cli-level=DEBUG
...
-------------------------------------------- Captured stderr call ---------------------------------------------
  File "/home/rmcgover/src/pubtools-pulp/tests/command.py", line 91, in test
    fn()
  File "/home/rmcgover/src/pubtools-pulp/pubtools/_pulp/task.py", line 164, in main
    self.run()
  File "/home/rmcgover/src/pubtools-pulp/pubtools/_pulp/tasks/clear_repo.py", line 135, in run
    publish_fs = self.publish_with_cache_flush(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rmcgover/src/pubtools-pulp/pubtools/_pulp/tasks/common.py", line 187, in publish_with_cache_flush
    out = self.flush_cdn(repos)
          ^^^^^^^^^^^^^^^^^^^^^
  File "/home/rmcgover/src/pubtools-pulp/pubtools/_pulp/step.py", line 37, in new_fn
    ret = fn(instance, *args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rmcgover/src/pubtools-pulp/pubtools/_pulp/tasks/common.py", line 78, in flush_cdn
    return [purge_repo(r) for r in repos if r.relative_url]
                                            ^^^^^^^^^^^^^^
AttributeError: 'MapFuture' object has no attribute 'relative_url'
```

This is a baseline update that you should not have accepted.
To update the baselines, you run the tests with `UPDATE_BASELINES=1`, review the result and then add it to your commit only if the new result looks right, which it doesn't in this case.

The root cause is type confusion again.  Unfortunately it looks like there's quite a mess here now.  I see that there's two `publish` methods, one expects a `list[Repository]` and the other a `list[Future[Repository]]`.  This particular code expects to be seeing `Repository` and not `Future[Repository]`, but we already determined that other nearby code paths are expecting to see futures.

So: I think this code has been inadvertently written in such a way that it only works if the given objects can be treated as *both* `Future[Repository]` and `Repository`.

That sucks, but anyway an easy fix would be:

```
--- a/pubtools/_pulp/tasks/clear_repo.py
+++ b/pubtools/_pulp/tasks/clear_repo.py
@@ -2,7 +2,7 @@ import logging
 from functools import partial
 
 import attr
-from more_executors.futures import f_map, f_sequence
+from more_executors.futures import f_map, f_proxy, f_sequence
 from pubtools.pulplib import ContainerImageRepository, Criteria
 
 from pubtools._pulp.arguments import SplitAndExtend
@@ -133,7 +133,7 @@ class ClearRepo(
 
         # They should be published.
         publish_fs = self.publish_with_cache_flush(
-            [f_map(f, lambda cr: cr.repo) for f in cleared_repos_fs]
+            [f_proxy(f_map(f, lambda cr: cr.repo)) for f in cleared_repos_fs]
         )
 
         # Wait for everything to finish
```

...which allows the code to continue being sloppy.

I think it'd be better if all this code was reviewed and type hints added to resolve the confusion throughout, but I don't think we can expect that to fall into the scope of your current issue.

lekang9: yes i did notice that there was a separate error but i could not figure out why, it just felt that the inputs were not right but you have answered it as above

lekang9: i'll go ahead and revert baseline update from the last commit and then update it to use f_proxy

lekang9: so i've reverted the previous commit and changed the input to what you gave me above and there was only baseline updates related to "set_cdn_published" which looks alot better so i went ahead and did that but i think that the skipping log can be handled better:
```
>           raise AssertionError(message)
E           AssertionError: Output differs from expected (set UPDATE_BASELINES=1 if intended):
E           --- <output of test>
E           +++ /home/lekang/Repos/pubtools-pulp/tests/logs/clear_repo/test_clear_repo/test_clear_file_skip_publish.txt
E           @@ -6,10 +6,9 @@
E            [    INFO] Record push items: started
E            [    INFO] Record push items: finished
E            [    INFO] Publish: skipped
E           +[    INFO] Set cdn_published: started
E           +[    INFO] Set cdn_published: finished
E            [    INFO] Flush UD cache: started
E            [    INFO] UD cache flush is not enabled.
E            [    INFO] Flush UD cache: finished
E           -[    INFO] Flush CDN cache: started
E           -[    INFO] CDN cache flush is not enabled.
E           -[    INFO] Flush CDN cache: finished
```
in relation to how flushing of cdn cache is flushed when publish is skipped, its simply skipped but there is no info logging there. going to look into that i think i should make proper use of the step decorator to skip the flush_cdn step so that it does what it would for publish (if that makes sense)

lekang9: This looks better then there being no logs
```
>           raise AssertionError(message)
E           AssertionError: Output differs from expected (set UPDATE_BASELINES=1 if intended):
E           --- <output of test>
E           +++ /home/lekang/Repos/pubtools-pulp/tests/logs/clear_repo/test_clear_repo/test_clear_file_skip_publish.txt
E           @@ -6,6 +6,9 @@
E            [    INFO] Record push items: started
E            [    INFO] Record push items: finished
E            [    INFO] Publish: skipped
E           +[    INFO] Flush CDN cache: started
E           +[    INFO] CDN cache flush is not enabled.
E           +[    INFO] Flush CDN cache: finished
E            [    INFO] Set cdn_published: started
E            [    INFO] Set cdn_published: finished
E            [    INFO] Flush UD cache: started

tests/command.py:208: AssertionError
```

lekang9: I tried to address some of the complaints in static that seemed simpler to address
```
static create: /home/lekang/Repos/pubtools-pulp/.tox/static
static installdeps: -rtest-requirements.txt
static inst: /home/lekang/Repos/pubtools-pulp/.tox/.tmp/package/1/pubtools-pulp-1.27.1.zip
static installed: alabaster==0.7.13,astroid==3.0.1,attrs==23.1.0,Babel==2.13.1,bandit==1.7.5,black==23.11.0,certifi==2023.11.17,cffi==1.16.0,cfgv==3.4.0,charset-normalizer==3.3.2,click==8.1.7,coverage==6.5.0,coveralls==3.3.1,cryptography==41.0.5,decorator==5.1.1,dill==0.3.7,distlib==0.3.7,docopt==0.6.2,docutils==0.20.1,edgegrid-python==1.3.1,exceptiongroup==1.2.0,fastpurge==1.0.4,filelock==3.13.1,frozendict==2.3.9,frozenlist2==1.0.0,gitdb==4.0.11,GitPython==3.1.40,gssapi==1.8.3,humanize==4.9.0,identify==2.5.32,idna==3.6,imagesize==1.4.1,importlib-metadata==6.8.0,importlib-resources==6.1.1,iniconfig==2.0.0,iniparse==0.5,isort==5.12.0,Jinja2==3.1.2,jsonschema==4.20.0,jsonschema-specifications==2023.11.1,kobo==0.34.0,koji==1.33.1,markdown-it-py==3.0.0,MarkupSafe==2.1.3,mccabe==0.7.0,mdurl==0.1.2,mock==5.1.0,more-executors==2.11.4,mypy-extensions==1.0.0,ndg-httpsclient==0.5.1,nodeenv==1.8.0,packaging==23.2,pathspec==0.11.2,pbr==6.0.0,pkgutil_resolve_name==1.3.10,platformdirs==4.0.0,pluggy==1.3.0,pre-commit==3.5.0,pubtools==1.3.0,pubtools-pulp @ file:///home/lekang/Repos/pubtools-pulp/.tox/.tmp/package/1/pubtools-pulp-1.27.1.zip,pubtools-pulplib==2.38.0,pushcollector==1.3.0,pushsource==2.40.0,pyasn1==0.5.1,pycparser==2.21,Pygments==2.17.2,pylint==3.0.2,pyOpenSSL==23.3.0,pytest==7.4.3,pytest-cov==4.1.0,pytest-threadleak==0.5.0,python-dateutil==2.8.2,pytz==2023.3.post1,PyYAML==6.0.1,referencing==0.31.0,requests==2.31.0,requests-gssapi==1.2.3,requests-mock==1.11.0,requests-toolbelt==1.0.0,rhsm==1.19.2,rich==13.7.0,rpds-py==0.13.1,rpm==4.17.1,rpm-py-installer==1.2.0,six==1.16.0,smmap==5.0.1,snowballstemmer==2.2.0,Sphinx==7.1.2,sphinx-argparse==0.4.0,sphinxcontrib-applehelp==1.0.4,sphinxcontrib-devhelp==1.0.2,sphinxcontrib-htmlhelp==2.0.1,sphinxcontrib-jsmath==1.0.1,sphinxcontrib-qthelp==1.0.3,sphinxcontrib-serializinghtml==1.1.5,stevedore==5.1.0,tomli==2.0.1,tomlkit==0.12.3,typing_extensions==4.8.0,urllib3==2.1.0,virtualenv==20.24.7,zipp==3.17.0
static run-test-pre: PYTHONHASHSEED='2065978990'
static run-test: commands[0] | sh -c 'pylint pubtools; test $(( $? & (1|2|4|32) )) = 0'
************* Module pubtools._pulp.tasks.copy_repo
pubtools/_pulp/tasks/copy_repo.py:52:37: R1729: Use a generator instead 'any(not r_id.strip() for r_id in pair)' (use-a-generator)
************* Module pubtools._pulp.tasks.common
pubtools/_pulp/tasks/common.py:210:4: W0246: Useless parent or super() delegation in method 'add_args' (useless-parent-delegation)
************* Module pubtools._pulp.tasks.delete
pubtools/_pulp/tasks/delete.py:405:21: R1721: Unnecessary use of a comprehension, use list(adv_f.result()) instead. (unnecessary-comprehension)
************* Module pubtools._pulp.tasks.fix_cves
pubtools/_pulp/tasks/fix_cves.py:36:18: R1721: Unnecessary use of a comprehension, use list(erratum_page) instead. (unnecessary-comprehension)
pubtools/_pulp/tasks/fix_cves.py:92:15: R1721: Unnecessary use of a comprehension, use list(resp.result()) instead. (unnecessary-comprehension)
************* Module pubtools._pulp.tasks.push.items.base
pubtools/_pulp/tasks/push/items/base.py:126:23: R1718: Consider using a set comprehension (consider-using-set-comprehension)
************* Module pubtools._pulp.tasks.set_maintenance.set_maintenance_on
pubtools/_pulp/tasks/set_maintenance/set_maintenance_on.py:40:26: R1718: Consider using a set comprehension (consider-using-set-comprehension)

------------------------------------------------------------------
Your code has been rated at 9.98/10 (previous run: 9.94/10, +0.03)

ERROR: InvocationError for command /usr/bin/sh -c 'pylint pubtools; test $(( $? & (1|2|4|32) )) = 0' (exited with code 1)
```
i'm not sure why it still gives an error does it do so until all complaints are addressed?

lekang9: Rajul helped to point out which specific complaint was causing static to fail, it was confusing because it started with a W i assumed it was a warning rather than an error to fail on

pubtools/_pulp/tasks/common.py
```diff
@@ -179,7 +182,11 @@ def publish_with_cache_flush(
         # hook implementation(s) may now flush pulp-derived caches and datastores
         pm.hook.task_pulp_flush()
 
-        # flush CDN cache
+        # if publish is being skipped then cdn flush should also be skipped
+        if "publish" in self.args.skip and "flush_cdn" not in self.args.skip:
```
rohanpm, line 186: Minor complaint: I think this kind of logic would be better moved to where the arguments are initialized, e.g. in PulpTask.args method.

The reason is that you're adding a side-effect to publish_with_cache_flush here: the "args" may now be different after the method returns.  It's not easy for other developers to be aware that this is happening.  If someone writes code elsewhere which checks whether self.args.skip contains "flush_cdn", it'll behave differently depending on whether that code runs before or after this method, making it bug-prone.

Whereas if you moved this initialization into PulpTask.args, then the value is always consistent.  Everything else should only read from `args` and never write to it, making it easier to understand.

tests/logs/copy_repo/test_copy_repo/test_copy_container_repo.jsonl
```diff
@@ -0,0 +1,2 @@
+{"event": {"type": "check-repos-start"}}
```
rohanpm, line 1: Why does your pull request show the introduction of all these `test_copy_repo` files?

You've probably pulled in some content from a PR that Nathan made while you were working on this. This is not ideal as the history will show both of you introducing these files, while they were really introduced by him.

Could you please rebase this PR and squash it into a single commit on top of the latest master?  That should make these go away.

pubtools/_pulp/tasks/common.py
lekang9: @rohanpm  i've been thinking about what you suggested here but is it possible if the arguments haven't yet been parsed within PulpTask? (unless im mistaken)

prior to making any of these changes i had originally thought of making the change within step.py in the StepDecorator class (which was abandoned fairly early)

i believe it was something similar to
```
    def should_skip(self, instance):
        skip = (getattr(instance.args, "skip", None) or "").split(",")
        if "publish" in skip and not "flush_cdn" in skip:
            skip.append("flush_cdn")
        return self.machine_name in skip
```

but i guess above doesn't make sense because each step is independent of each other aside from having the same skip arguments
Maybe something like
```
    def should_skip(self, instance):
        skip = getattr(instance.args, "skip", None) or ""
        if self.machine_name == "flush_cdn" and "publish" in skip and "flush_cdn" not in skip:
            skip.append("flush_cdn")
        return self.machine_name in skip
```

i ran above through the tests and it seems to be passing the tests, i suppose we could maintain some kind of conditional_skip function where we could hold these and apply it within should_skip function
it ended up looking like:
```
    def conditional_skip(self, skip):
        cond_skip = skip
        if self.machine_name == "flush_cdn" and "publish" in cond_skip and "flush_cdn" not in cond_skip:
            cond_skip.append("flush_cdn")
        return cond_skip

    def should_skip(self, instance):
        skip = getattr(instance.args, "skip", None) or ""
        skip = self.conditional_skip(skip)
        return self.machine_name in skip
```
it seems to check off even if code is written elsewhere if im not mistaken and also other code is not reliant on publish_with_cache_flush so i suppose we could undo a ton of things we've done and still include the skip without changing code elsewhere

lekang9: out of curiosity i went ahead and cloned down latest master and applied above, the tests passed successfully (unexpected) i expected it to fail on baselines because flush_cdn should be skipped in addition to publish changing the baselines but i guess i was wrong?

lekang9: i pushed up the corresponding changes related to what i'm referring to above on top of the latest master, let me know what you think

thanks

pubtools/_pulp/step.py
```diff
@@ -51,8 +51,19 @@ def new_fn(instance, *args, **kwargs):
 
         return new_fn
 
+    def conditional_skip(self, skip):
+        cond_skip = skip
+        if (
+            self.machine_name == "flush_cdn"
```
rohanpm, line 57: This isn't the right value.  The step is declared as:

```
@step("Flush CDN cache")
```

That makes the machine name: "flush_cdn_cache"

So this never matches, which is why the branch shows as uncovered.

Overall, while I like the approach of handling it in the decorator, I think it shouldn't be hardcoded for just this case.  My ideal solution would look like this:

```
@step("Flush CDN cache", depends_on=["publish"])
```

The meaning being: this step depends on those steps, so if those steps were skipped, then this one should be too.  If implemented once correctly it could easily be applied to any steps with a similar relationship in the future.

pubtools/_pulp/tasks/common.py
rohanpm: > is it possible if the arguments haven't yet been parsed within PulpTask? 

I'm not quite sure what you mean by this; `self.args` is a property using lazy initialization, the args are going to be parsed the first time they're requested (and never parsed again after that). The suggestion of moving some of the arg handling logic into that property would have no impact on when they get initialized, so if this is not a problem with the initialization right now then it wouldn't be a problem after making that change either.

pubtools/_pulp/step.py
lekang9: i've understand what needs to happen but i'm having a hard time understanding how to have the step decorator accept this additional argument so that it can be used in the StepDecorator class

rohanpm: Not tested, but it should be something like this.

```
diff --git a/pubtools/_pulp/step.py b/pubtools/_pulp/step.py
index b606a01..0783ac2 100644
--- a/pubtools/_pulp/step.py
+++ b/pubtools/_pulp/step.py
@@ -9,8 +9,9 @@ LOG = logging.getLogger("pubtools.pulp")
 class StepDecorator(object):
     """Implementation of PulpTask.step decorator. See that method for more info."""
 
-    def __init__(self, name):
+    def __init__(self, name, depends_on):
         self._name = name
+        self._depends_on = depends_on or []
 
     @property
     def human_name(self):
@@ -53,7 +54,10 @@ class StepDecorator(object):
 
     def should_skip(self, instance):
         skip = (getattr(instance.args, "skip", None) or "").split(",")
-        return self.machine_name in skip
+        # We might be *explicitly* skipped,
+        # or *implicitly* skipped if any of our deps are skipped
+        deps_and_self = self._depends_on + [self.machine_name]
+        return any([name in skip for name in deps_and_self])
 
 
 # helpers used in implementation of decorator
diff --git a/pubtools/_pulp/task.py b/pubtools/_pulp/task.py
index 8e7dc92..6e00e74 100644
--- a/pubtools/_pulp/task.py
+++ b/pubtools/_pulp/task.py
@@ -68,7 +68,7 @@ class PulpTask(object):
         return self._args
 
     @classmethod
-    def step(cls, name):
+    def step(cls, name, depends_on=None):
         """A decorator to mark an instance method as a discrete workflow step.
 
         Marking a method as a step has effects:
@@ -94,7 +94,7 @@ class PulpTask(object):
         - The step is considered *failed* if it raised an exception.
         - The step is considered *finished* once all items have been yielded.
         """
-        return StepDecorator(name)
+        return StepDecorator(name, depends_on)
 
     def _basic_args(self):
         # minimum args required for a pulp CLI task
diff --git a/pubtools/_pulp/tasks/common.py b/pubtools/_pulp/tasks/common.py
index 9aed2d8..553c977 100644
--- a/pubtools/_pulp/tasks/common.py
+++ b/pubtools/_pulp/tasks/common.py
@@ -36,7 +36,7 @@ step = PulpTask.step
 class CDNCache(FastPurgeClientService, CdnClientService):
     """Provide features to interact with CDN cache."""
 
-    @step("Flush CDN cache")
+    @step("Flush CDN cache", depends_on=["publish"])
     def flush_cdn(self, repos):
         """Clears the CDN cache for the repositories provided
```

lekang9: thanks, that helps visualize it alot i had not adjusted PulpTask only the StepDecorator class in step.py

lekang9: also i had a much less elegant solution with if conditional inside of should_skip, the way you are showing it here is much cleaner

lekang9:
```
        # They should have UD cache flushed.
        to_await.extend(self.flush_ud(repos))
    
        # They should have CDN cache flushed.
        to_await.extend(self.flush_cdn(repos))
    
        # Now make sure we wait for everything to finish.
        for f in to_await:
>           f.result()
E           AttributeError: 'FileRepository' object has no attribute 'result'

pubtools/_pulp/tasks/clear_repo.py:151: AttributeError
```
after some poking around it looks like the tests are telling me clear_repo.py doesn't check if there is no result since cdn cache flush was skipped so it seems like i'll need to do something there to ensure that if skip value is supplied then it shouldn't try to get at that attribute or there is some baseline default result so that any skipped steps would have the attribute still

lekang9: i'm trying to understand:
a skipped function is decorated and is as follows:
```
        def new_fn(instance, *args, **kwargs):
            if self.should_skip(instance):
                LOG.info(
                    "%s: skipped",
                    self.human_name,
                    extra={"event": {"type": "%s-skip" % self.machine_name}},
                )
                return args[0] if args else None
```
and so since flush_cdn is being skipped now based on the skipped publish the return value appears to be None and only the ud flush result is being printed

i remembered as i typed this out about the pdb flag
```
>           f.result()
E           AttributeError: 'FileRepository' object has no attribute 'result'

pubtools/_pulp/tasks/clear_repo.py:151: AttributeError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/lekang/Repos/pubtools-pulp/pubtools/_pulp/tasks/clear_repo.py(151)run()
-> f.result()
(Pdb) print(f)
FileRepository(id='some-filerepo', eng_product_id=123, relative_url='some/publish/url', signing_keys=[], is_sigstore=False, mutable_urls=[])
```
i'm not sure exactly what is happening at that point, so i went to take a look at to_await which is the result of record_push_items which seems to be connected to pushsource i looked at it briefly but i don't think i need to go so deep into that to resolve this

rohanpm: It's type confusion about futures vs non-futures again.  The problem is that `flush_cdn` normally returns Futures resolved when the flush has completed, but if the step is skipped then it instead returns `args[0]` which is just a list of repositories rather than futures.

Here's a solution I'd consider for this.  In summary, when defining a step, also make it possible to define what value should be returned if the step is skipped.  In this case it'd make sense to return an empty list (because we did 0 cache flushes in total).

```diff
diff --git a/pubtools/_pulp/step.py b/pubtools/_pulp/step.py
index 3d372d9..7d0f9df 100644
--- a/pubtools/_pulp/step.py
+++ b/pubtools/_pulp/step.py
@@ -4,14 +4,16 @@ import inspect
 from more_executors.futures import f_sequence, f_return
 
 LOG = logging.getLogger("pubtools.pulp")
+UNSET = object()
 
 
 class StepDecorator(object):
     """Implementation of PulpTask.step decorator. See that method for more info."""
 
-    def __init__(self, name, depends_on):
+    def __init__(self, name, depends_on, skipped_value):
         self._name = name
         self._depends_on = depends_on or []
+        self._skipped_value = skipped_value
 
     @property
     def human_name(self):
@@ -29,7 +31,11 @@ class StepDecorator(object):
                     self.human_name,
                     extra={"event": {"type": "%s-skip" % self.machine_name}},
                 )
-                return args[0] if args else None
+                return (
+                    self._skipped_value
+                    if self._skipped_value is not UNSET
+                    else args[0] if args else None
+                )
 
             logger = StepLogger(self)
             args = logger.log_start(args)
diff --git a/pubtools/_pulp/task.py b/pubtools/_pulp/task.py
index 4437ed1..f8f528a 100644
--- a/pubtools/_pulp/task.py
+++ b/pubtools/_pulp/task.py
@@ -4,7 +4,7 @@ from argparse import ArgumentParser, RawDescriptionHelpFormatter
 
 from pubtools.pluggy import task_context
 
-from .step import StepDecorator
+from .step import StepDecorator, UNSET
 
 
 LOG = logging.getLogger("pubtools.pulp")
@@ -68,7 +68,7 @@ class PulpTask(object):
         return self._args
 
     @classmethod
-    def step(cls, name, depends_on=None):
+    def step(cls, name, depends_on=None, skipped_value=UNSET):
         """A decorator to mark an instance method as a discrete workflow step.
 
         Marking a method as a step has effects:
@@ -77,6 +77,8 @@ class PulpTask(object):
         - The method can be skipped if requested by the caller (via --skip argument)
         - Methods that depend on other methods are implicitly skipped if depends_on
             is supplied and that dependant is being skipped
+        - If the method is skipped, it returns either `skipped_value` (if that has
+          been set), or the method's first argument.
 
         Steps may be written as plain blocking functions, as non-blocking
         functions which accept or return Futures, or as generators.
@@ -96,7 +98,7 @@ class PulpTask(object):
         - The step is considered *failed* if it raised an exception.
         - The step is considered *finished* once all items have been yielded.
         """
-        return StepDecorator(name, depends_on)
+        return StepDecorator(name, depends_on, skipped_value)
 
     def _basic_args(self):
         # minimum args required for a pulp CLI task
diff --git a/pubtools/_pulp/tasks/common.py b/pubtools/_pulp/tasks/common.py
index 553c977..cc7d235 100644
--- a/pubtools/_pulp/tasks/common.py
+++ b/pubtools/_pulp/tasks/common.py
@@ -36,7 +36,7 @@ step = PulpTask.step
 class CDNCache(FastPurgeClientService, CdnClientService):
     """Provide features to interact with CDN cache."""
 
-    @step("Flush CDN cache", depends_on=["publish"])
+    @step("Flush CDN cache", depends_on=["publish"], skipped_value=[])
     def flush_cdn(self, repos):
         """Clears the CDN cache for the repositories provided
```

lekang9: that makes alot of sense now, thank you

- flush_cdn accepts a list of repository as the argument
- flush_cdn gets implicitly skipped
- args[0] is this list
- without the above solution then the return value is unexpected when normally flush_cdn returns a list of futures
- when we try to access the result of the future via f.result it is like trying to get at method result from the futures but we are doing so on the repository which it tries to do anyways but result is not an attribute of that object so it complains

it was confusing at first but i think the gist of it has been mostly cleared up

lekang9: It looks like there is a loss of coverage by the change indirectly https://app.codecov.io/gh/release-engineering/pubtools-pulp/pull/266/blob/pubtools/_pulp/services/fakepulp.py but i'm not sure how my changes might have affected that coverage unless i pulled something in by mistake

rohanpm: @lekang9 for the purpose of this PR you can just ignore that.  There are two codecov checks, codecov/project and codecov/patch.  codecov/patch is at 100% for your change which means *you* didn't pull down the coverage.

Coverage was somehow decreased during a recent dependency update, it shouldn't block this PR.

PR #267, opened 2024-03-18
Fix crashes on YAML serialization after dependency update

rohanpm:
Tests started to crash with:

    yaml.representer.RepresenterError: cannot represent an object

... on fields using frozendict or frozenlist.
I'm not exactly sure what changed to introduce this error, as PyYAML has not been updated recently.

It can be resolved by explicitly telling the YAML dumper how to serialize those types.

### DIFF
--- a/pubtools/_pulp/services/fakepulp.py
+++ b/pubtools/_pulp/services/fakepulp.py
@@ -3,6 +3,8 @@ import logging
 
 import yaml
 import attr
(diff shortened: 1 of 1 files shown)


rohanpm:
This is supposed to fix the CI failures seen in dependency update https://github.com/release-engineering/pubtools-pulp/pull/259 .

PR #271, opened 2024-04-03
Update upload_repo for arc split [RHELDST-23264]

amcmahon-rh:
Currently, all rpm and errata content is pushed to all-rpm-content repo in each environment. Accumulation in the repos is generating concern for performance. To alleviate the build-up, we'll be splitting content between several repos; 'all-rpm-content-XX' for rpms and 'all-erratum-content-YYYY' for errata.

This change refactors the upload_repo and upload_context to reflect this split. It does not include validation for the errata year range, as that feels like a specific implimentation detail.

### DIFF
--- a/pubtools/_pulp/tasks/push/items/base.py
+++ b/pubtools/_pulp/tasks/push/items/base.py
@@ -95,6 +95,24 @@ class PulpPushItem(object):
     to a single unit (e.g. modulemd YAML files; comps.xml files).
     """
 
+    MULTI_UPLOAD_CONTEXT = False
+    """
+    Can the push item class create a different upload context based on the data
+    it holds. e.g PulpRpmPushItem has checksum separated upload repos/contexts.
+    
+    If true, subclass should override upload_repo.
+    """
+
+    @property
+    def upload_repo(self):
+        """Given a PulpPushItem, returns the repo its pushsource.PushItem
+        should be uploaded to.
+
+        Subclasses MAY override this to return a variable default destination.
+        Subclasses must set MULTI_UPLOAD_CONTEXT to True for this to be used.
+        """
+        return "" # pragma: no cover
+
     @classmethod
     def for_item(cls, pushsource_item, **kwargs):
         """Given a pushsource.PushItem, returns an instance of a PulpPushItem wrapper
@@ -136,12 +154,11 @@ class PulpPushItem(object):
 
         return klass.match_items_units(items, units)
 
-    @classmethod
-    def upload_context(cls, pulp_client):
+    def upload_context(self, pulp_client):
         """Return a context object used during uploads.
 
-        The context object will be shared across all uploads for a specific content
-        type.
+        The context object should be shared across all uploads for a specific
+        content type. Subclasses classes may introduce multiple contexts.
 
         Subclasses MAY override this to provide their own context
         (e.g. to cache a value rather than recalculating per upload).
--- a/pubtools/_pulp/tasks/push/items/base.py
+++ b/pubtools/_pulp/tasks/push/items/base.py
@@ -111,7 +111,7 @@ class PulpPushItem(object):
         Subclasses MAY override this to return a variable default destination.
         Subclasses must set MULTI_UPLOAD_CONTEXT to True for this to be used.
         """
-        return "" # pragma: no cover
+        return ""  # pragma: no cover
 
     @classmethod
     def for_item(cls, pushsource_item, **kwargs):
--- a/pubtools/_pulp/tasks/push/items/erratum.py
+++ b/pubtools/_pulp/tasks/push/items/erratum.py
@@ -47,13 +47,17 @@ class PulpErratumPushItem(PulpPushItem):
         # is formatted like RHXA-YYYY
         name_match = self.ADVISORY_PATTERN.match(self.pushsource_item.name)
         if not name_match:
-            LOG.error("Bad Advisory name: '%s' does not contain a reasonable year value.",
-                         self.pushsource_item.name)
+            LOG.error(
+                "Bad Advisory name: '%s' does not contain a reasonable year value.",
+                self.pushsource_item.name,
+            )
             raise ErratumPushItemException
         year = int(name_match.group(1))
         if not any([r[0] <= year <= r[1] for r in self.CONTENT_SPLIT_RANGES]):
-            LOG.warning("%s was not in a valid date range for repo content splitting, using the default.",
-                        self.pushsource_item.name )
+            LOG.warning(
+                "%s was not in a valid date range for repo content splitting, using the default.",
+                self.pushsource_item.name,
+            )
             year = "0000"
         return "all-erratum-content-%s" % year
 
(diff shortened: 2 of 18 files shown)


amcmahon-rh:
@negillett ptal too

negillett:
- coverage, lgtm

pubtools/_pulp/tasks/push/items/erratum.py
```diff
@@ -29,13 +29,17 @@ class ErratumUploadContext(UploadContext):
...
+    MULTI_UPLOAD_CONTEXT = True
+
+    @property
+    def upload_repo(self):
+        # Split errata into different repos by year, assuming the advisory name
+        # is formatted like RHXA-YYYY
+        try:
+            return "all-erratum-content-%s" % self.pushsource_item.name[5:9]
+        except KeyError:
+            # Poorly formatted errata name, use default
```
rohanpm, line 41: This seems like dead code.  Are you sure this can be reached?

If we assume that `pushsource_item.name` at least is a `str`, I don't think we can get here.

Firstly, if you index a string (or list) out of bounds, you don't get KeyError, you get IndexError.

Secondly, if you *slice* a string (or list) out of bounds, you don't get an error at all, you get an incomplete (or empty) sequence of the same type.

```
>>> x = 'abc'

>>> x[5]
  File "<stdin>", line 1, in <module>
IndexError: string index out of range

>>> x[1:9]
'bc'

>>> x[5:9]
''
```

pubtools/_pulp/tasks/push/items/erratum.py
```diff
@@ -29,13 +29,17 @@ class ErratumUploadContext(UploadContext):
...
-    # For more details, refer RHELDST-12782.
-    UPLOAD_REPO = "all-rpm-content"
+    MULTI_UPLOAD_CONTEXT = True
+
+    @property
+    def upload_repo(self):
+        # Split errata into different repos by year, assuming the advisory name
+        # is formatted like RHXA-YYYY
+        try:
+            return "all-erratum-content-%s" % self.pushsource_item.name[5:9]
```
rohanpm, line 39: I think we should at least have some minimal validation of this, not necessarily to handle it, but at least to make it crash ASAP if our expectation is not met. Can we please do that?

For example, just wrapping it with `int(self.pushsource_item.name[5:9])` to crash if the string portion is not numeric would be good enough for me.  Then an unexpected advisory name would cause it to crash early and immediately draw attention to this code.

If it's not validated at all, an unexpected advisory name can cause it to try uploading to a nonexistent `all-erratum-content-something` repo.  A developer looking at an error like that is pretty likely to waste time thinking that the error is that the repo is expected to exist but doesn't, and it may take them a while to figure out that the real problem was that this code received an unexpected advisory name.

rohanpm (changes requested):

pubtools/_pulp/tasks/push/items/erratum.py
amcmahon-rh: Rohan, the more I think about it, the more I feel like pubtools-pulp should check the year ranges. It's an implementation detail specific to our version of pulp, but without some checking we wont end up using the default repo at all. Do we know how likely it is we'll be dealing with errata older than 2014?

pubtools/_pulp/tasks/push/items/erratum.py
```diff
@@ -18,24 +18,50 @@
...
-    # Erratums are always uploaded to this repo first.
-    # Pulp stores the erratum pkglist for (erratum_id, repo) pair. If an erratum is
-    # uploaded for an already existing ID, the pkglist is overwritten if the repo is
-    # same, else is appended. To keep the pkglist consistent, all the erratums are
-    # uploaded to the same repo instead of a random repo from the list of destinations.
-    # For more details, refer RHELDST-12782.
-    UPLOAD_REPO = "all-rpm-content"
+    MULTI_UPLOAD_CONTEXT = True
+    ADVISORY_PATTERN = re.compile(r"RH.A-(\d{4})", re.IGNORECASE)
+    CONTENT_SPLIT_RANGES = [(2014, 2030), (8014, 8030)]
```
rohanpm, line 44: The other side of the code is being updated to use 2040 as the end of the range.

pubtools/_pulp/tasks/push/items/erratum.py
rohanpm: I think that's pretty unlikely.  I think 2016 was the birth of the current way of doing things, so when I mentioned 2014 as the earliest year I was already adding a couple of extra years to be safe.

pubtools/_pulp/tasks/push/items/base.py
```diff
@@ -95,6 +95,24 @@ class PulpPushItem(object):
...
+
+    @property
+    def upload_repo(self):
+        """Given a PulpPushItem, returns the repo its pushsource.PushItem
+        should be uploaded to.
+
+        Subclasses MAY override this to return a variable default destination.
+        Subclasses must set MULTI_UPLOAD_CONTEXT to True for this to be used.
+        """
+        return ""  # pragma: no cover
```
negillett, line 114: should this instead raise not implemented?

pubtools/_pulp/tasks/push/phase/upload.py
```diff
@@ -60,13 +60,20 @@ def run(self):
...
-                ctx = upload_context[item_type]
-
+                if item.MULTI_UPLOAD_CONTEXT:
+                    if item_type not in upload_context:
+                        upload_context[item_type] = {}
+                    if item.upload_repo not in upload_context[item_type]:
+                        upload_context[item_type][item.upload_repo] = (
+                            item.upload_context(self.pulp_client)
+                        )
+                    ctx = upload_context[item_type][item.upload_repo]
```
negillett, line 70: this is going to raise key error if the condition immediately above isn't met

rohanpm (commented):
What's here looks good to me;

However I won't approve it yet because the referenced repos don't seem to exist. Since this new behavior is used all the time with no way to disable it, the all-erratum-content-* repos should exist in all relevant environments *before* submitting this, otherwise submitting this will put pubtools-pulp master into an unreleaseable state.

tests/push/test_pulperratumpushitem.py
```diff
@@ -30,3 +35,31 @@ def test_erratum_publishes_all_repos():
...
+    # Test upload_repo maps name to default repo when advisory year is outside
+    # the expected range
+
+    item = PulpErratumPushItem(pushsource_item=ErratumPushItem(name="RHSA-1999:1234"))
+
+    assert item.upload_repo == "all-erratum-content-0000"
+
+
+def test_erratum_upload_repo_bad_format():
+    # Test upload_repo maps name to default repo when advisory year is outside
```
rohanpm, line 58: This comment doesn't match what's actually happening.

amcmahon-rh:
Rajul is looking into the test coverage failures. It appears to be some new rate limiting issue. Running coverage locally, all new lines have been covered. The new rhsm-tools has been released, but we can this PR open til I've confirmed the new repos have been created. Talking with Peter / Michal about it now.

rajulkumar (approved):

rajulkumar:
Required repos were created on QA, Stage and Prod. Merging this now.

PR #280, opened 2024-04-18
Fix CopyRepo task [RHELDST-23737, RHELDST-23709] 

rbikar:
Most notable fixes and changes:
* added support for all content types - because of that we need to use
  different `Criteria` for `copy_content()` call. We can't use
 `Criteria.with_unit_type()` for all types because we don't currently support
  class->content_type mapping for all content types in pubtools-pulplib. 
  So there is a hybrid approach for content types supporting `with_unit_type`
  `Criteria` and those types that don't support that where we use `with_field`
  variant. 
* previously when unsupported content type was passed, `copy_content()`
  was called with `criteria=None` which lead to copying of all content
  from repo + flawed `if` conditions called `copy_content()` multiple
  time (one extra call for each unknown content type)
* now we use one `copy_content()` call for one repo pair but because there can
   be spawned more that one pulp task for we split criteria
* where supported we use `unit_fields` for `Criteria` in order to
   lower RAM requirements
* all repo copies are properly aggregated in one `Copy content` step.
* added test logs for tests

### DIFF
--- a/pubtools/_pulp/tasks/copy_repo.py
+++ b/pubtools/_pulp/tasks/copy_repo.py
@@ -1,16 +1,22 @@
 import logging
+from collections import namedtuple
 from functools import partial
+from itertools import chain
 
 import attr
-from more_executors.futures import f_map, f_sequence
+from more_executors.futures import f_map, f_sequence, f_proxy
 from pubtools.pulplib import (
     ContainerImageRepository,
     Criteria,
-    ErratumUnit,
+    Matcher,
     FileUnit,
-    ModulemdUnit,
     RpmUnit,
+    ErratumUnit,
+    ModulemdUnit,
+    ModulemdDefaultsUnit,
+    YumRepoMetadataFileUnit,
 )
+
 
 from pubtools._pulp.arguments import SplitAndExtend
 from pubtools._pulp.services import CollectorService, PulpClientService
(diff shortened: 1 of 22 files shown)


rbikar:
Notes:
* recording of push items is currently incomplete and supports only a few types, I am not currently unsure e.g. how to handle Comps types - which are aggregated in one pushsource type but in copy repo we treat them as separate 4 types. This may need some further improvements.
* @MichalHaluza  we talked about possibility of throttling the number of parallel repo copies, I believe we can achieve that via `--pulp-throttle PULP_THROTTLE` option, but it'll need a little change in rcm-pub args handling for copy-repos command.

rohanpm (approved):
Is it thought that the memory usage problem was mainly down to many copy tasks unexpectedly copying all the units, so they'd be loaded into memory several times?

I had thought that you might need to specify the unit fields used during the copy (which the old code does), otherwise it might load all of the fields including some of the very heavy fields on RPM units.

MichalHaluza (dismissed):

rbikar:
@rohanpm 
Well, definitely adding the unit_fields for RPM and SRPM units would be very prudent.

We think that the previous implementation drained resources on Pub and I guess the performance was negatively affected  (on pulp side) by not using unit_fields at least for S/RPM units. I'll have another into this.

rbikar:
@rohanpm  @MichalHaluza  ptal.
I have improved the solution a bit, and now the queries to pulp are limited by `unit_fields` where supported.
Unfortunately this has  a little trade-off - we can't currently combine `Criteria` with more than one `content_type_id`  when using `with_unit_type` criteria (but I think it should be allowed and pulp supports it, pubtools-pulplib complains with too complicated `Criteria`), so for each repo pair there will be more copy tasks spawned but  I believe it shouldn't have significant performance issues.

MichalHaluza (approved):

rohanpm (approved):

rbikar:
squashed commits and rebased -> merginng.

PR #284, opened 2024-05-07
Load ProductID certificate metadata via pushsource [RHELDST-24276]

MichalHaluza:
python-rhsm cannot be installed with GCC 14. Given that project has been deprecated for years, it's unlikely the actual underlying issue will be resolved. Pubtools-pulp uses python-rhsm to read ProductID certificates and pull product metadata from them. This commit leverages code for ProductID reading newly introduced in Pushsource instead of python-rhsm.

This MR also replaces **rpm-py-installer** with **rpmdyn** in the same fashion like it was done in https://github.com/release-engineering/pushsource/commit/1c1b4f92ba104f1c8b5c644256881e713def98d2


MichalHaluza:
This MR depends on https://github.com/release-engineering/pushsource/pull/476

pubtools/_pulp/tasks/push/items/productid.py
```diff
@@ -80,8 +79,7 @@ class PulpProductIdPushItem(PulpDirectUploadPushItem):
     @product_versions.default
     def _product_versions_from_cert(self):
         if self.pushsource_item and self.pushsource_item.src:
```
rohanpm, line 81: I think this shouldn't check `src` any more.  Now that there is a proper attribute for `products`, it should just be trusted and we shouldn't look at other attributes on the item to decide whether or not to use the `products` attribute.

tests/push/test_productid_updates_versions.py
```diff
@@ -94,7 +87,10 @@ def test_updates_product_versions(monkeypatch, tmpdir):
     # make an item targeting two of the repos
     item = PulpProductIdPushItem(
         pushsource_item=ProductIdPushItem(
-            name="test", src=str(productid), dest=["repo2", "repo7"]
+            name="test",
+            src=str(productid),
+            dest=["repo2", "repo7"],
+            products=FakeCert(["a", "d"]),
```
rohanpm, line 93: Using a `FakeCert` as a value for `products` attribute doesn't seem to make sense.

I think these Fake classes are just obsolete now, shouldn't it be something like this using the class added into pushsource:

```
products=[Product(version="a"), Product(version="d")]
```

rbikar (approved):

tests/push/test_productid_updates_versions.py
```diff
@@ -79,12 +68,6 @@ def test_updates_product_versions(monkeypatch, tmpdir):
...
     ctrl.insert_repository(repo8)
 
-    # Set up cert parser to use a fake cert object (saves us having to explicitly
-    # generate certs with certain values)
-    monkeypatch.setattr(
-        rhsm.certificate, "create_from_file", lambda _: FakeCert(["a", "d"])
-    )
-
     # make a fake productid.
     # content doesn't matter since we patched the cert parser, it just has
```
rohanpm, line 72: Minor: this block could also have been deleted.  The comment is now false and the block is creating an empty file which I believe is never accessed.

rohanpm (dismissed):

tests/push/test_productid_updates_versions.py
MichalHaluza: Looks like it's still needed as `pubtools.pulplib._impl.fake.client.FakeClient._do_upload_file()` [opens](https://github.com/release-engineering/pubtools-pulplib/blob/master/pubtools/pulplib/_impl/fake/client.py#L386) that file for reading and fails if it doesn't exists.

rohanpm: I see.  I was right about the comment being false at least (there is no patching of a cert parser).  I leave it to you to decide how much to care about the comment's accuracy.

rohanpm (dismissed):

tests/push/test_productid_updates_versions.py
```diff
@@ -79,22 +68,22 @@ def test_updates_product_versions(monkeypatch, tmpdir):
...
 
-    # Set up cert parser to use a fake cert object (saves us having to explicitly
-    # generate certs with certain values)
-    monkeypatch.setattr(
-        rhsm.certificate, "create_from_file", lambda _: FakeCert(["a", "d"])
-    )
-
     # make a fake productid.
-    # content doesn't matter since we patched the cert parser, it just has
+    # content doesn't matter since we mock the ProductIDs, it just has
```
rohanpm, line 72: Err... sorry to be pedantic again about this comment (but you have said in other reviews that you like it).

I don't think "mock" is the right word here, and I think using it contributes to confusion around what mocking is.  Well, I don't think it's a very well-defined term, but I agree with Wikipedia's leading sentence: "A mock object is an object that imitates a production object in limited ways."  You are providing *real* ProductId objects here, which are going to work in exactly the same way as ProductId objects loaded out of a cert.  You're using the API of pushsource in a normal and valid way and I don't see how it should be considered "mocking".  You're just providing your own data which skips the default behavior of trying to load that data from a file.

If I may give an analogy, if there was a function with a signature like:

```python
def load_config(path: str = "/etc/some-default.cfg"):
    ...
```

...and some code were to call that function like `cfg = load_config("my-non-default.cfg")`, I don't think that should be called "mocking" the path or the config?

I believe it's better for the comment to say "we provide our own ProductIds", and also for us to think about the code in this way.  Or if you wanted to emphasize that this is overriding the default method of obtaining the objects, you could even use the term "inject" per [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection), "a programming technique in which an object or function receives other objects or functions that it requires, as opposed to creating them internally".

rohanpm (dismissed):

rbikar (approved):

rohanpm (approved):

PR #287, opened 2024-06-04
Don't flush UD cache for repos without eng ID [RHELDST-24551]

negillett:


negillett:
@rajulkumar @rohanpm, could you ptal when you have a moment?

rajulkumar (approved):

tests/publish/test_publish.py
```diff
@@ -238,6 +255,46 @@ def test_repo_publish_cache_cleanup(command_tester):
...
+
+    # pulp repo is published
+    assert [hist.repository.id for hist in fake_pulp.publish_history] == ["repo4"]
+    # flushed the urls
+    assert sorted(fake_publish.fastpurge_client.purged_urls) == [
+        "https://cdn.example.com/content/unit/4/client/mutable1",
+        "https://cdn.example.com/content/unit/4/client/mutable2",
+    ]
+    # did not flush the UD object
+    assert fake_publish.udcache_client.flushed_repos == []
```
rajulkumar, line 295: Good to have this test case include a repo with eng ID as well so that it covers the all the cases i.e. this or any change in future makes the code skip only the relevant repo without an eng ID only

negillett: Hmm. Yeah, could do that. I almost included this new repo in an existing test to achieve that effect. Probably helpful to see both in a single case, as you say, so I'll add.

rajulkumar (approved):

MichalHaluza (approved):

PR #288, opened 2024-06-05
Fix erratum unit upload [RHELDST-24497]

rbikar:
After we introduced erratum unit upload to special repos like `all-erratum-content-XXXX`, unwanted behavior came up: erratum unit previously uploaded to `all-rpm-content` was uploaded to different upload repository which caused extending its pkglist.

This behavior is not wanted as we need to overwrite the pkglist. In order to keep wanted behaviour, the already existing erratum units in `all-rpm-content` are uploaded to the same repository.


rbikar:
cc @amcmahon-rh   @MichalHaluza  for awareness

rajulkumar (approved):

MichalHaluza (approved):

PR #302, opened 2024-08-13
Fix: Publish.publish_with_cache_flush with skip publish [RHELDST-25040]

negillett:
Previously, if publish_with_cache_flush was called and skip of publish was requested, more_executors' f_sequence would raise a type error. This is because publish returns the first given argument when skipped (list of repo objects), not a future as f_sequence requires.

This commit adds an empty list as the publish step's "skipped_value", replacing the repo list argument return value.


tests/delete/test_delete_packages.py
```diff
@@ -1,20 +1,18 @@
...
+    Matcher,
+    ModulemdUnit,
+    RpmUnit,
+    YumRepository,
 )
 
-from pubtools._pulp.ud import UdCacheClient
-
 from pubtools._pulp.tasks.delete import Delete, entry_point
+from pubtools._pulp.ud import UdCacheClient
```
negillett, line 15: by isort

src/pubtools/_pulp/tasks/common.py
```diff
@@ -119,9 +120,15 @@ def publish_with_cache_flush(
         # publish the repos found
         publish_fs = self.publish(repos)
 
-        # wait for the publish to complete before
-        # flushing caches.
-        f_sequence(publish_fs).result()
+        if publish_fs and all(isinstance(f, Future) for f in publish_fs):
```
rohanpm, line 123: This fix feels quite ugly and risky.  Risky because it can hide type errors, and because skipping await of publish in cases where we should not is potentially very severe (e.g. exodus-gw commit will be attempted before pulp tasks have finished).

A very similar issue already came up for the `flush_cdn` method (which has since been deleted).  I think we should fix it the same way it was fixed there, in https://github.com/release-engineering/pubtools-pulp/pull/266 .

Essentially you do:

```diff
--- a/src/pubtools/_pulp/tasks/common.py
+++ b/src/pubtools/_pulp/tasks/common.py
@@ -70,7 +70,7 @@ class Publisher(UdCache):
             action="store_true",
         )
 
-    @step("Publish")
+    @step("Publish", skipped_value=[])
     def publish(self, repos):
         out = []
```

This tells the `@step` decorator "if the step is skipped, the returned value should be an empty list".  An empty list represents that zero publishes occurred.  All the existing code here should just work, as it's entirely valid to do `f_sequence(publish_fs).result()` on an empty list.

rohanpm (changes requested):

src/pubtools/_pulp/tasks/common.py
negillett: Thanks, I'll give that a try

negillett: Looks good!

rohanpm (approved):

PR #314, opened 2024-10-18
Add --all-repos argument to deletion task [RHELDST-13644]

amcmahon-rh:
There are occasions where we would like to remove a package from all repos its associated with. This can be tedious as every repo has to be  listed out. This change adds a convenient --all-repos argument to address this issue.


src/pubtools/_pulp/tasks/delete.py
```diff
@@ -564,26 +577,39 @@ def map_to_repo(self, units, repos, unit_attr):
...
+                        )
+                    else:
+                        repo_map.setdefault(repo, []).append(unit)
+                        unit_map.get(unit_name).repos.append(repo)
+
+            missing = set(repos) - set(repo_map.keys())
+            if missing:
+                missing = ", ".join(sorted(list(missing)))
+                LOG.warning("No units to remove from %s", missing)
+        else:
```
rbikar, line 599: Is this else branch really needed? It duplicates a few lines from the if branch. Can this somehow be combined? It is still deleting content from repos - only repos are calculated in a different way.

src/pubtools/_pulp/tasks/delete.py
```diff
@@ -564,26 +577,39 @@ def map_to_repo(self, units, repos, unit_attr):
...
+            LOG.info("Get repos from units")
+            LOG.info(list(units))
+            for unit in sorted(units):
+                unit_name = getattr(unit, unit_attr)
+                unit_map.setdefault(unit_name, RemoveUnitItem(unit=unit, repos=[]))
+                for repo in unit.repository_memberships:
+                    LOG.info(repo)
+                    if re.match("all-rpm-content-.*", repo):
```
rbikar, line 607: Not sure if we need to do this. If user runs delete command for `--all-repos`, they'll probably want to delete it really from all repos.

rbikar (commented):
Added a few suggestions, ptal @amcmahon-rh .

src/pubtools/_pulp/tasks/delete.py
rajulkumar: AFAIR packages were usually deleted from the associated repos that were published to the users, unless explicitly defined to clear them from all repos including all-rpm-content.

As discussed earlier, I think it would be good to keep it this way. However, the user should be able to remove the content from all-rpm-content when provided with --repo. 

Wrt to the above comment to get rid of this else block and determine the repos from the membership if none provided, this case to skip the all-rpm-content with --all-repos and include when provided with --repo might look a bit messy. Hence, probably relengs could provide a better insight on the heuristics whether to include all-rpm-content in --all-repos.

Also, "all-rpm-content.*" instead of "all-rpm-content-.* , otherwise it will skip "all-rpm-content" repo.

rajulkumar: @MichalHaluza ^

rajulkumar (changes requested):

src/pubtools/_pulp/tasks/delete.py
```diff
@@ -78,10 +78,19 @@ def add_args(self):
...
+            help="remove content from these comma-seperated repositories. "
+                 "This argument is mutually exclusive with --all-repos.",
             type=str,
             action=SplitAndExtend,
             split_on=",",
+            default=[]
+        )
+
+        self.parser.add_argument(
+            "--all-repos",
```
rohanpm, line 90: I see that this is aligned with what the JIRA asked for, but I personally feel making it support `--repo *` might have been better than `--all-repos`.

The main reason for this is that there are several "layers" here (pub client => hub => worker => pubtools-pulp), and the value of "*" could be simply passed through the layers the same as any other repo ID and then interpreted by the final command as "every repo".  All the related complexity would live only in a single layer.

With `--all-repos` and `--repo` being separate and mutually exclusive options though, it seems more likely that multiple layers are going to have to learn about the new option and add new logic to deal with it.

src/pubtools/_pulp/tasks/delete.py
rohanpm: I'm in favor of keeping this logic and not removing it from `all-rpm-content-.*` even with `--all-repos`.

Existing "delete" commands only have the postcondition that a given piece of content becomes inaccessible from the customer's point of view, not necessarily implying that content has been truly deleted from the remote services.  Deleting from `all-rpm-content-.*` definitely isn't needed in order to achieve that postcondition.

src/pubtools/_pulp/tasks/delete.py
rohanpm: Additionally, per later thread about "special" all-rpm-content repos:

If this were implemented as `--repo *` with `*` meaning "all customer repos containing the content", then the user has the choice of passing something like `--repo * --repo all-rpm-content-ab --repo all-rpm-content-cd [...]` to set up a single task which removes files from all customer repos *and also* all-rpm-content repos, if that's what they want to do.

This seems to be impossible if we make `--repo` and `--all-repos` two mutually exclusive options (or potentially leads us to more complexity, like adding a third boolean option to toggle whether `--all-repos` includes special repos).

rohanpm (dismissed):

src/pubtools/_pulp/tasks/delete.py
```diff
@@ -78,7 +78,9 @@ def add_args(self):
 
         self.parser.add_argument(
             "--repo",
-            help="remove content from these comma-seperated repositories ",
+            help="remove content from these comma-seperated repositories. "
+                 "If '*' is used, the packge will be removed from all repos,"
```
rohanpm, line 82:
```suggestion
                 "If '*' is used, the package will be removed from all repos, "
```

src/pubtools/_pulp/tasks/delete.py
```diff
@@ -564,21 +566,29 @@ def map_to_repo(self, units, repos, unit_attr):
...
             unit_map.setdefault(unit_name, RemoveUnitItem(unit=unit, repos=[]))
-            for repo in repos:
-                if repo not in unit.repository_memberships:
-                    LOG.warning(
-                        "%s is not present in %s",
-                        unit_name,
-                        repo,
-                    )
-                else:
+            if "*" in repos:
```
rohanpm, line 572: Minor: not really liking the fact that this goes into two separate branches.
It means, for example, that if someone later refactors the two lines starting from `repo_map.setdefault(repo, []).append(unit)` they will need to refactor both branches or will possibly waste time forgetting to update one of them.

It seems like (in psuedocode) it could all be handled like this:

```
# note: repos probably needs to be copied before this, if the caller doesn't expect
# it to be mutated.
if repos contains the wildcard:
  update repos, replacing the wildcard with all repos containing the unit
process all repos
```

Right now the "process all repos" bit appears in the code twice.

tests/logs/delete/test_delete_packages/test_delete_rpms_with_all_repos.txt
```diff
@@ -0,0 +1,37 @@
...
+[ WARNING] Nothing mapped for removal
+[    INFO] Unassociate files: finished
+[    INFO] Record push items: started
+[    INFO] Record push items: finished
+[    INFO] Delete files: finished
+[    INFO] Delete RPMs: started
+[    INFO] Get RPMs: started
+[    INFO] 2 unit(s) found for deletion
+[    INFO] Get RPMs: finished
+[ WARNING] No units to remove from *
```
rohanpm, line 17: The warning seems undesirable.  I think it's going to be produced every time `*` is used even if content is actually being removed?

This problem might also be avoided if the code worked more like the psuedo-code from the earlier thread.

src/pubtools/_pulp/tasks/delete.py
```diff
@@ -78,7 +79,10 @@ def add_args(self):
 
         self.parser.add_argument(
             "--repo",
-            help="remove content from these comma-seperated repositories ",
+            help="remove content from these comma-seperated repositories. "
+                 "If '%s' is used, the package will be removed from all repos,"
```
rohanpm, line 83: Looks to be missing a space:

```suggestion
                 "If '%s' is used, the package will be removed from all repos, "
```

Without this it will look not quite right, like this:

`the package will be removed from all repos,excluding all-rpm-content-*`

rohanpm (dismissed):

src/pubtools/_pulp/tasks/delete.py
```diff
@@ -563,8 +567,17 @@ def log_missing_units(self, searched_units, unit_attr, unit_type, unit_names):
     def map_to_repo(self, units, repos, unit_attr):
         repo_map = {}
         unit_map = {}
+        if ALL_REPOS_INDICATOR in repos:
+            repos = repos.copy()
+            for unit in sorted(units):
```
rohanpm, line 572: I'm just realizing the negative impact of expanding `*` here, once, covering all units - rather than expanding it every time, *per unit*, in the later loop.

That seems to mean `*`, rather than meaning "Every repo containing the unit being deleted", means "Every repo containing *any* unit in the set of units being deleted".

It doesn't quite seem right, but it doesn't seem like it will majorly break anything either.  It means there will be requests to delete units from repos which already don't contain the unit.  Pointless but not particularly harmful.

Does it produce a bunch of warning spam though, i.e. the "(unit) is not present in (repo)" log in the upcoming loop?

The scenario would be something like:

- unit_a is in repo_a
- unit_b is in repo_b

If user does `delete unit_a --repo *`, followed by `delete unit_b --repo *`, everything is fine.
But if user does `delete unit_a unit_b --repo *`, then it will warn "unit_a is not present in repo_b" and "unit_b is not present in repo_a", won't it?  That does not seem to warrant a warning.

If we want `*` to have the meaning "every repo containing the unit being deleted", then I don't see a way to avoid expanding the repo list separately for each unit.

tests/logs/delete/test_delete_packages/test_delete_rpms_with_all_repos.txt
```diff
@@ -0,0 +1,38 @@
...
+[    INFO] Unassociate files: finished
+[    INFO] Record push items: started
+[    INFO] Record push items: finished
+[    INFO] Delete files: finished
+[    INFO] Delete RPMs: started
+[    INFO] Get RPMs: started
+[    INFO] 2 unit(s) found for deletion
+[    INFO] Get RPMs: finished
+[ WARNING] dash-2.25-1.test8_x86_64.rpm is not present in all-rpm-content-xp
+[ WARNING] dash-2.25-1.test8_x86_64.rpm is not present in other-yumrepo
```
rohanpm, line 18: I guess this is showing what I'm saying above.

Why is this warning warranted?  The user did not provide any input implying that dash-2.25-1.test8_x86_64.rpm was expected to be present in other-yumrepo.

rohanpm (approved):

PR #324, opened 2024-12-04
Use pushitem content to upload

rajulkumar:
pushitem content method added in latest pushsource update returns a file-like object with the item bits that may be used for upload. This decouples the source of the items and expectation to exist as a file on local fs.

Note: This PR will be applicable after the changes to `pushsource` are merged
https://github.com/release-engineering/pushsource/pull/643


rohanpm:
Not a big deal, but technically `requirements.in` should be updated to depend on the minimum version of pushsource providing this new API.

rohanpm (approved):

PR #327, opened 2024-12-10
copy-repo makes only two copy calls for a repo pair [RHELDST-28235]

rajulkumar:
Copy-repo previously made copy calls for each content type. Lots of Pulp tasks were spawned with this creating a big queue and slowing down the operations. Hence, this updates copy-repo to make only two copy calls, one for rpm and other for non-rpm content types.

Non-rpm content type units are smaller in count and unit_fields are smaller in size, hence they are all clubbed in the same call ignoring the unit_fields.

Also, non-rpm content type units are copied first as they may have modulemd units, which should be copied before the corresponding modular rpm units to avoid those rpms being visible to user without modulemd units in case of failure or partial copy.

### DIFF
--- a/src/pubtools/_pulp/tasks/copy_repo.py
+++ b/src/pubtools/_pulp/tasks/copy_repo.py
@@ -4,7 +4,7 @@ from functools import partial
 from itertools import chain
 
 import attr
-from more_executors.futures import f_map, f_sequence, f_proxy
+from more_executors.futures import f_map, f_sequence, f_proxy, f_flat_map, f_return
 from pubtools.pulplib import (
     ContainerImageRepository,
     Criteria,
@@ -73,6 +73,23 @@ CONTENT_TYPES = (
     ContentType(("package_langpacks",)),
 )
 
+RPM_CONTENT_TYPES = (
+    "rpm",
+    "srpm",
+)
+
+NON_RPM_CONTENT_TYPES = (
+    "iso",
+    "erratum",
+    "modulemd",
+    "modulemd_defaults",
+    "yum_repo_metadata_file",
+    "package_group",
+    "package_category",
+    "package_environment",
+    "package_langpacks",
+)
+
 
 @attr.s(slots=True)
 class RepoCopy(object):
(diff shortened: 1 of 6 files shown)


rbikar:
@rajulkumar 

I don't think that currently the code guarantee that modulemd copies (non-rpm) are finished before rpm copies. Can you have look, please? If I get it right, the copies are only enqueued in correct order by order of execution can be different. 
But in reality we haven't that ensured even with the original approach as this command isn't run that frequently we didn't get any complaints about that.

rajulkumar:
@rbikar 
I updated the copy_content method to ensure the criteria are processed and completed sequentially for repo pair.
Also, pylint gave "cell-var-from-loop" warning, so checked the units in the controller for the updates to repository memberships wrt to src and dest repos and it looked fine. Hence, disabled the warning there.

negillett (approved):

src/pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -214,14 +249,23 @@ def repo_copy(copy_tasks, repo):
...
+        for src_repo, dest_repo in sorted(repo_pairs):
             one_pair_copies = []
+            tasks_f = f_return()
             for item in criteria or [None]:
-                tasks_f = self.pulp_client.copy_content(
-                    src_repo, dest_repo, criteria=item
+                # ensure the criterias are processed and completed/resolved in order
+                # so that non-rpm copy completes before rpm copy
+                # pylint:disable=cell-var-from-loop
+                tasks_f = f_flat_map(
```
rbikar, line 259: yeah, I think we should be fine with this pylint warning, the issue is with those variables that change each loop and are used in lambda. And if you use it like this you will got all lambdas called with values of the last iteration:
```
fcs = []
for x in range(10):
    f = lambda: x
    fcs.append(f)

for func in fcs:
    print(func())

"""
OUTPUT:
9
9
9
9
9
9
9
9
9
9
"""
```
But iiuic we should be fine because we call it immediately and test didn't discover any issue with it.

rbikar (approved):

PR #333, opened 2025-01-07
Fix copy_repo sending incorrect repo pairs when providing multiple pairs

rajulkumar:
Copy_repo waits for the complete copy of non-rpm content before copying rpm content. However, there were some instances where previous repo pair was sent for non-rpm content copy when multiple repo pairs were provided.

Lambda functions seems to be the reason for the error as they resolve the values during execution rather than creation, hence might be picking incorrect values. This patch replaces it with a modified partial function method that resolves the values during creation instead of execution.

### DIFF
--- a/src/pubtools/_pulp/tasks/copy_repo.py
+++ b/src/pubtools/_pulp/tasks/copy_repo.py
@@ -102,6 +102,14 @@ class RepoCopy(object):
     """The repo to which content was copied."""
 
 
+class freeze_arguments(partial):
+    # execute the function only with the args and kwargs that were
+    # provided while creating the object. args and kwargs provided
+    # during the call are ignored.
+    def __call__(self, /, *args, **keywords):
+        return self.func(*self.args, **self.keywords)
+
+
 class CopyRepo(CollectorService, PulpClientService, PulpRepositoryOperation):
     @property
     def content_type_criteria(self):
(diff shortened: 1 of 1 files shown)


src/pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -102,6 +102,14 @@ class RepoCopy(object):
     """The repo to which content was copied."""
 
 
+class partpartial(partial):
```
rohanpm, line 105: Minor: I am not really fond of the name, because to me it implies a *weaker* kind of partial ("part partial"), while this is actually *stronger* than a usual partial object. It freezes *all* the arguments, while I think partial's name comes from the fact that it's typically used to *partially* freeze/bind a function's arguments.

Not a big deal obviously, given it's not public API.

Some other ideas for names: `freeze_arguments`, `bind_arguments`, `full_partial`.

rohanpm (approved):

src/pubtools/_pulp/tasks/copy_repo.py
rajulkumar: Thanks for the insight. 
"Part partial" came from the fact that it's not providing all the features of a "partial", hence the name. However, suggested names sounds more meaningful.

rohanpm (approved):

PR #339, opened 2025-01-21
Add a docstring for copy-repo [RHELDST-28512]

Gdetrane:
Add docstring to the CopyRepo class for completing the docs generation.

### DIFF
--- a/src/pubtools/_pulp/tasks/copy_repo.py
+++ b/src/pubtools/_pulp/tasks/copy_repo.py
@@ -118,6 +118,7 @@ class CopyRepo(CollectorService, PulpClientService, PulpRepositoryOperation):
     Copied content may optionally be filtered on content types.
     If the user provides a non existing repo, the command fails.
     """
+
     @property
     def content_type_criteria(self):
         # Only return non-None if there were really any types given.
--- a/src/pubtools/_pulp/tasks/copy_repo.py
+++ b/src/pubtools/_pulp/tasks/copy_repo.py
@@ -111,6 +111,13 @@ class freeze_arguments(partial):
 
 
 class CopyRepo(CollectorService, PulpClientService, PulpRepositoryOperation):
+    """
+    Copy content from one repository to another.
+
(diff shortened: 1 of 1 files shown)


src/pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -111,6 +111,12 @@ def __call__(self, /, *args, **keywords):
 
 
 class CopyRepo(CollectorService, PulpClientService, PulpRepositoryOperation):
+    """
+    Copy content from one repository to another.
+
+    This command copies RPM and SRPM content from repository repo-A to repo-B. If the user provides a non existing repo, the command fails.
```
rohanpm, line 117: Why does it say "RPM and SRPM content"?  It works for other content types as well, doesn't it?

Gdetrane: I'll keep it more general then

src/pubtools/_pulp/tasks/copy_repo.py
```diff
@@ -111,6 +111,12 @@ def __call__(self, /, *args, **keywords):
 
 
 class CopyRepo(CollectorService, PulpClientService, PulpRepositoryOperation):
+    """
+    Copy content from one repository to another.
+
+    This command copies content from repository repo-A to repo-B. If the user provides a non existing repo, the command fails.
```
rajulkumar, line 117: Multiple repo pairs could be provided in the same command. Also, the content to be copied can be filtered on the content types. So we should include theses specifics too.
This could be something like

> This command copies the content between the provided source,destination repository pairs. Copied content may optionally be filtered on content types. 
If the user provides a non existing repo, the command fails.

rajulkumar (changes requested):

src/pubtools/_pulp/tasks/copy_repo.py
Gdetrane: I explain it in the detailed description itself, so I might as well add it to the docstring, it's not immediately clear you can have multiple pairs in a single command, I'll change this immediately

PR #345, opened 2025-03-06
Remove EngID flush [RHELDST-29701]

drepelov:
Previously, during each Flush UD cache step, the Eng ID for each flushed repository was also flushed. This resulted in repeated flushes of the same Eng ID.
It turns out that flushing the Eng ID is not necessary at all, and that just flushing the repos should be enough to keep the UD up to date.

### DIFF
--- a/src/pubtools/_pulp/ud.py
+++ b/src/pubtools/_pulp/ud.py
@@ -105,7 +105,9 @@ class UdCacheClient(object):
             Future[None]
                 A future resolved once flush has completed.
         """
-        warnings.warn("`flush_product()` function will be deprecated.", DeprecationWarning)
+        warnings.warn(
+            "`flush_product()` function will be deprecated.", DeprecationWarning
+        )
         return self._flush_object("eng-product", product_id)
 
     def flush_repo(self, repo_id):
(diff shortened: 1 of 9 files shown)


src/pubtools/_pulp/ud.py
```diff
@@ -92,19 +92,6 @@ def _flush_object(self, object_type, object_id):
 
         return out
 
-    def flush_product(self, product_id):
```
rbikar, line 95: I'm just wondering, if we really need to remove this from the `UdCacheClient` - although it won't be problematic in pubtools-pulp but it basically is backwards incompatible change and we should avoid doing such changes. 
So I would keep it at least in this place.  

There were also plans to implement sth like `pubtools-pulp flush-cache` where flushing by product id may be useful but because of Konflux it got very low priority and plans were dropped. 

You could also add deprecation notice and create a JIRA that would remove this piece of code in a year or so.

rbikar (changes requested):

src/pubtools/_pulp/ud.py
drepelov: Here is the jira for the future retirement https://issues.redhat.com/browse/RHELDST-30255

rbikar (approved):

PR #348, opened 2025-03-20
Do not record push items for copy repo task [RHELDST-30557]

rbikar:
* in typical workflows the number of push items in copy repo task is very high (100K+ items) and it extremely overloads downstream systems (Pub) and makes this copy task unusable so let's now skip the push items recording as there is also very little value in storing all those push items.


rbikar:
@MichalHaluza fyi

MichalHaluza (approved):

PR #403, opened 2026-06-05
Disallow RPM upload with the same NVR [RHELDST-37824]

rbikar:
This change introduces a functionality that will raise a fatal error
when upload of an RPM with identical `cdn_path` of already present unit in pulp
is attempted.

The RPMs in question differ only with their checksums (and signing keys),
further association to live repositories and publish will cause breakage
of repository.

Duplicates check is by default disabled, and can be enabled by setting
PUBTOOLS_PULP_ALLOW_DUPLICATE_UNITS env var.

This should be a workaround until proper suport for such scenario is
implemented.

### DIFF
--- a/src/pubtools/_pulp/tasks/push/items/rpm.py
+++ b/src/pubtools/_pulp/tasks/push/items/rpm.py
@@ -151,16 +151,25 @@ class PulpRpmPushItem(PulpPushItem):
         Publishing such units would break integrity of repositories when published becuase of collision on origin path.
         """
         crit = Criteria.and_(
-            Criteria.with_unit_type(RpmUnit, unit_fields=["filename", "cdn_path", "sha256sum"]),
+            Criteria.with_unit_type(
+                RpmUnit, unit_fields=["filename", "cdn_path", "sha256sum"]
+            ),
             Criteria.with_field("filename", self.pushsource_item.name),
         )
 
         results = pulp_client.search_content(crit)
         for item in results:
-            if item.cdn_path and item.cdn_path == self.cdn_path and item.sha256sum != self.pushsource_item.sha256sum:
-                msg = "Fatal error: Duplicate RPM present in Pulp: %s, sha256: %s, cdn_path: %s" % (
-                item.filename,
-                item.sha256sum,
-                item.cdn_path,
-            )
+            if (
+                item.cdn_path
+                and item.cdn_path == self.cdn_path
+                and item.sha256sum != self.pushsource_item.sha256sum
+            ):
+                msg = (
+                    "Fatal error: Duplicate RPM present in Pulp: %s, sha256: %s, cdn_path: %s"
+                    % (
+                        item.filename,
+                        item.sha256sum,
+                        item.cdn_path,
+                    )
+                )
                 raise RuntimeError(msg)
--- a/src/pubtools/_pulp/tasks/push/items/rpm.py
+++ b/src/pubtools/_pulp/tasks/push/items/rpm.py
@@ -144,3 +144,23 @@ class PulpRpmPushItem(PulpPushItem):
 
     def upload_to_repo(self, repo):
         return repo.upload_rpm(self.pushsource_item.content(), cdn_path=self.cdn_path)
+
+    def fail_if_duplicate(self, pulp_client):
+        """
+        If there is already an RPM with identical CDN path of requested within push present in Pulp but has different checksum, raise an error.
+        Publishing such units would break integrity of repositories when published becuase of collision on origin path.
+        """
+        crit = Criteria.and_(
+            Criteria.with_unit_type(RpmUnit, unit_fields=["filename", "cdn_path", "sha256sum"]),
+            Criteria.with_field("filename", self.pushsource_item.name),
+        )
+
+        results = pulp_client.search_content(crit)
+        for item in results:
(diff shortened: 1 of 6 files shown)


src/pubtools/_pulp/tasks/push/items/rpm.py
```diff
@@ -144,3 +149,31 @@ def ensure_uploaded(self, ctx, repo_f=None):
...
         return repo.upload_rpm(self.pushsource_item.content(), cdn_path=self.cdn_path)
+
+    def fail_if_duplicate(self, pulp_client):
+        """
+        Check for duplicate units in Pulp. If duplicate is found, raise an error.
+
+        This is a workaround until we have a proper support for units with same NVR but different checksum in Pulp.
+        Such units are not allowed to be uploaded to Pulp as they would break the integrity of the repository when published.
+        """
+        # import pdb; pdb.set_trace()
```
MichalHaluza: Leftover from debugging

MichalHaluza (commented):
Has this been tested on production-like data? AFAIK, we used to have the same RPMs signed with different GPG keys for different products in the past. This was (and still is) perfectly valid as long as the RPMs with the same NVRA but different signature aren't associated to the same repo. Checking at the time of upload may break these cases.

rbikar:
> Has this been tested on production-like data? AFAIK, we used to have the same RPMs signed with different GPG keys for different products in the past. This was (and still is) perfectly valid as long as the RPMs with the same NVRA but different signature aren't associated to the same repo. Checking at the time of upload may break these cases.

Haven't done any big tests on production data, but the case you mentioned will be disabled with current solution.
I just wanted workflow to fail as soon as possible.

I can move this check  to `Associate` phase probably where I'll do the same or similar query but directly to `dest repo` ,
maybe (even better) doing query only to `dest repo` would be sufficient even in the `Upload` phase. Now we do generic content query.

With new query we will enable upload of NVR duplicates unless the same NVR is present in `dest` repo. Sounds good?

MichalHaluza:
Yes, checking only the dest repo would be preferred, however can we actually do that? Are the dest repos already known during the upload step? (Note that the same RPM can be uploaded to multiple repos during a signle push, and in RHEL it oftentimes is - mainline and dot)

rbikar:
> Yes, checking only the dest repo would be preferred, however can we actually do that? Are the dest repos already known during the upload step? (Note that the same RPM can be uploaded to multiple repos during a signle push, and in RHEL it oftentimes is - mainline and dot)

Yep, dest repos are known after `Query pulp` step.

rbikar:
Implementation update - now the error is raised only in case when RPM would create a duplicate in destination repository.

This allows to upload of RPMs with same NVR and different gpg signing keys as long as RPMs target different repositories. Although I think this may also lead to problems with re-signed RPMs with the same path to origin even when targeting different repos. We would probably need to check `cdn_path` value or the gpg signing key.

src/pubtools/_pulp/tasks/push/items/rpm.py
```diff
@@ -49,16 +53,18 @@ def rpm_nvr(self):
             # mpr.hcraon.1.1.3_7le.41-0.4.4-slootnimda-api
             filename_rev = "".join(reversed(filename))
 
-            # 1.1.3_7le.41-0.4.4-slootnimda-api
-            nvr_rev = filename_rev.split(".", 2)[2]
+            # (hcraon, 1.1.3_7le.41-0.4.4-slootnimda-api)
```
MichalHaluza: Nitpick: All other comments below use apostrophes around string values. This one should too.

src/pubtools/_pulp/tasks/push/items/rpm.py
```diff
@@ -144,3 +150,36 @@ def ensure_uploaded(self, ctx, repo_f=None):
...
+
+    def fail_if_duplicate(self, pulp_client):
+        """
+        Check for duplicate units in destination repos of given push item. Should duplicate be found, raise an error.
+
+        This is a workaround until we have a proper support for units with same NVR but different checksum in Pulp.
+        Such units are not allowed to be uploaded to Pulp as they would break the integrity of the repository when published.
+        """
+        n, v, r, a = self.rpm_nvra
+        crit = Criteria.and_(
```
MichalHaluza, line 153: Just thinking out loud here - We could simplify the logic of the for-loop that evaluates the futures by adding a condition excluding the checksum of the item being uploaded here.

So if we're uploading `foo-1.2-3.noarch.rpm` with checksum `abc123`, the query would be `name=foo AND version=1.2 AND release=3 AND arch=noarch AND NOT checksum=abc123`. If the query doesn't return any results, it means that either the RPM with the same NVRA hasn't been uploaded to the repo, or it has been uploaded with the same checksum we're trying to reupload now, which would fall into the `continue` branch in the for-loop. If the query returns any number of results greater than 0, we know that the RPM with the same NVRA but different checksum is already in the repo and we don't even need to loop over the result and can `raise` immediately.

I'd _think_ this way would be more performant, but it's entirely possible that I'm wrong or the benefit is so small it's not even measurable.

rbikar:
After consultation with @MichalHaluza this should the correct solution:
* do not allow upload or association of RPM unit that has identical cdn_path (origin) as *any* unit in pulp.

src/pubtools/_pulp/tasks/push/items/rpm.py
rbikar: not relevant anymore

MichalHaluza (approved):

PR #406, opened 2026-07-23
WIP: Sync from external source - Konflux

rbikar:
* adds Sync phase that performs synchronization of a YUM repository from external source (pulp3) to internal Pulp.
* currently WIP as some details needs to be refined

Ref: RHELDST-38336, RHELDST-38337


rbikar:
@crungehottman 

this is still WIP but I managed to run this `Sync` phase on realistic environment almost successfully.
* we still have missing fix for sync in rhsm-pulp which disallowed to run whole push
* there are "couple" of TODOs that needs to be figured out (please have a look if you have time), maybe file some of them as follow-up JIRAs.
* I have concern about this part of pushsource code:  https://github.com/release-engineering/pushsource/blob/3befb6ad7e6b92a039efcf18c1ec990ea6653b6e/src/pushsource/_impl/backend/konflux_source/konflux_source.py#L308
** I think this maybe problematic for the new "multi key" signing - when they send us sth like "key_alias1,key_alias2" - this is currently not expected to be set on RPM units, here - https://github.com/release-engineering/pushsource/blob/master/src/pushsource/_impl/backend/koji_source.py#L345  it's read directly from RPM headers (this is not possible with Konflux source). Aliases won't work so we need to figure this out as well, maybe we can get this from pulp3 API, not sure right now. But please note that the signing keys handling will change after RHELDST-42989.
* tests TBD

crungehottman:
> @crungehottman
> 
> this is still WIP but I managed to run this `Sync` phase on realistic environment almost successfully.
> 
>     * we still have missing fix for sync in rhsm-pulp which disallowed to run whole push
> 

If you're testing on PiAAS, I built an image which contains the latest rhsm-pulp RPM (which contains the sync fix): `images.paas.redhat.com/crungeho/rhsm-pulp:latest`

Feel free to use that image in your PiAAS deployments.

>
>     * there are "couple" of TODOs that needs to be figured out (please have a look if you have time), maybe file some of them as follow-up JIRAs.
> 

It's not 100% clear to me which TODOs you intend to complete within the scope of this PR and which TODOs you prefer to be filed as Jira issues. I assume anything that would be required for functionality should be addressed within this PR, but any improvements (batching & multi-repo support, signature verification (the require_signature and allowed_keys TODOs)) can be filed as follow-up Jira issues (I'll file these issues in our epic, feel free to edit them as you see fit). That would leave the following to be addressed in this PR (I might be missing some context as to why they might need to be addressed in a follow-up Jira), but let me know what you think:

```
### only for sync phase?
### allow only erratum units to be uploaded for Konflux source?
TODO implement get_distribution() in pubtools-pulplib
TODO set notes.tmp: true
TODO skip all but rpm/srpm
TODO update cdn_path for each item, could be done in Update phase?
```

>     * I have concern about this part of pushsource code:  https://github.com/release-engineering/pushsource/blob/3befb6ad7e6b92a039efcf18c1ec990ea6653b6e/src/pushsource/_impl/backend/konflux_source/konflux_source.py#L308
>       ** I think this maybe problematic for the new "multi key" signing - when they send us sth like "key_alias1,key_alias2" - this is currently not expected to be set on RPM units, here - https://github.com/release-engineering/pushsource/blob/master/src/pushsource/_impl/backend/koji_source.py#L345  it's read directly from RPM headers (this is not possible with Konflux source). Aliases won't work so we need to figure this out as well, maybe we can get this from pulp3 API, not sure right now. But please note that the signing keys handling will change after RHELDST-42989.
> 

Good point, thanks for raising this concern. I have filed an issue to support multisig in KonfluxSource, 

>     * tests TBD

I assume the tests will be added in this PR?

src/pubtools/_pulp/services/pulp3.py
```diff
@@ -0,0 +1,77 @@
...
+        with self.__lock:
+            if not self.__instance:
+                self.__instance = self.new_pulp3_client()
+        return self.__instance
+
+    def new_pulp3_client(self, **kwargs):
+        """Creates and returns a new Pulp client with appropriate config."""
+        args = self._service_args
+
+        if not args.pulp3_url and not args.domain:
```
crungehottman: Should this be `or` rather than `and`? Currently this would allow one to be unset.

src/pubtools/_pulp/tasks/push/phase/sync.py
```diff
@@ -0,0 +1,129 @@
+import logging
+import uuid
+import anyio
+
+from .base import Phase
+import anyio
```
crungehottman: duplicate anyio import, was already imported on line 3

src/pubtools/_pulp/services/pulp3.py
```diff
@@ -0,0 +1,77 @@
...
+            LOG.error("Both pulp3-url and domain must be provided")
+            sys.exit(41)
+
+        # checks if pulp password is available as environment variable
+        if args.pulp3_user:
```
crungehottman, line 76: Should there be some error if pulp3_user isn't provided? Also, since auth is configured within this block, if a pulp3_user isn't provided, the return statement on line 67 would raise an UnboundLocalError because auth is never defined.

src/pubtools/_pulp/services/pulp3.py
```diff
@@ -0,0 +1,77 @@
...
+            pulp3_password = args.pulp3_password or os.environ.get("PULP3_PASSWORD")
+            if not pulp3_password:
+                LOG.warning("No pulp3 password provided for %s", args.pulp3_user)
```
crungehottman: Should this instead be `LOG.error` and sys.exit (similar behavior if pulp3-url and domain are undefined)?

src/pubtools/_pulp/tasks/push/phase/sync.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # PHASE II
+        # create repository in internal pulp to sync from ext source
+        # TODO support for multiple repos
+        # TODO chain this with futures
+        repo = self._create_tmp_repo()
+        self._sync_ext_repo(repo, dist_url)
+
+        for item in items_to_add:
+            # TODO update cdn_path for each item, could be done in Update phase?
```
crungehottman: I agree update phase makes sense as that is where the other cdn_path adjustments happen, the sync phase should probably just focus on getting the content into rhsm-pulp

src/pubtools/_pulp/tasks/push/command.py
```diff
@@ -113,6 +119,9 @@ def add_phase(klass, **kwargs):
                 update_push_items=collect_phase.in_queue.put,
                 publish_with_cache_flush=self.publish_with_cache_flush,
             )
+            if KONFLUX_SOURCE_ENABLED:
+                kwargs["pulp3_client"] = self.pulp3_client  ### only for sync phase?
```
crungehottman: I suppose if the cdn_path resolution takes place in the Update phase and will not require the pulp3 client, then the pulp3_client might only be needed in the sync phase? Or might it be needed elsewhere?

src/pubtools/_pulp/tasks/push/command.py
```diff
@@ -129,9 +138,15 @@ def add_phase(klass, **kwargs):
...
 
+        if KONFLUX_SOURCE_ENABLED:
+            # Sync RPMs from external source (e.g. Konflux) to Pulp
+            add_phase(Sync)
+
         # Ensure all items are uploaded to Pulp. This uploads bytes into Pulp
         # but does not guarantee the items are present in each of the desired
         # destination repos.
+
+        ### allow only erratum units to be uploaded for Konflux source?
```
crungehottman: Since the sync phase takes care of importing the (S)RPMs into rhsm-pulp, I believe Erratum units are the only units that should be uploaded (no need to upload RPMs nor SRPMs). Does that sound right?

rbikar: we can limit these only for Erratum units for now (maybe some other metadata files will be uploaded later)

src/pubtools/_pulp/tasks/push/command.py
rbikar: it's only for sync phase currently.

src/pubtools/_pulp/tasks/push/phase/sync.py
rbikar: actually I'm going to implement this on this place so we follow the "Upload phase" approach which sets cdn_path immediately after upload. 
We can also encapsulate this within pubtools-pulplib anb sth like `sync_with_update()`

src/pubtools/_pulp/services/pulp3.py
rbikar: yeah, `or` sounds better.

rbikar:
> > @crungehottman

> > * there are "couple" of TODOs that needs to be figured out (please have a look if you have time), maybe file some of them as follow-up JIRAs.
> > ```
> 
> It's not 100% clear to me which TODOs you intend to complete within the scope of this PR and which TODOs you prefer to be filed as Jira issues. I assume anything that would be required for functionality should be addressed within this PR, but any improvements (batching & multi-repo support, signature verification (the require_signature and allowed_keys TODOs)) can be filed as follow-up Jira issues (I'll file these issues in our epic, feel free to edit them as you see fit). That would leave the following to be addressed in this PR (I might be missing some context as to why they might need to be addressed in a follow-up Jira), but let me know what you think:

only the `cdn_path` is missing required functionality (+ one small thing in pulp3 client for getting distribution) + handling signing_keys not strictly required but for should have higher priority.


> I assume the tests will be added in this PR?

yes

rbikar:
@crungehottman 

sorry for delays, I'll address all these necessary issues asap after my pto.

rbikar:
@crungehottman 

This is now ready for review, ptal.
Please note that for this change to work, it needs https://github.com/release-engineering/pubtools-pulplib/pull/241/ for be merged and released.
Locally tests work fine, tested with live pulp3/rhsm-pulp.

.github/workflows/docs.yml
```diff
@@ -10,11 +10,11 @@ jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v7
```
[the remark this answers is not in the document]
rbikar: this affects multiple repos and I won't fix that here.

requirements.txt
```diff
@@ -1,18 +1,19 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.10
 # by the following command:
 #
-#    pip-compile --generate-hashes requirements.in
+#    pip-compile --generate-hashes --no-index requirements.in
```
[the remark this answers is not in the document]
rbikar: this also affects multiple repos and I won't fix that here and has been there for a long time.

.github/workflows/docs.yml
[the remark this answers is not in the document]
rbikar: no

requirements.txt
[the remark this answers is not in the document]
rbikar: no

src/pubtools/_pulp/services/pulp3.py
```diff
@@ -0,0 +1,109 @@
...
+        """Creates and returns a new Pulp3 client with appropriate config."""
+        args = self._service_args
+        auth = cert = None
+
+        if not (args.pulp3_url and args.domain):
+            LOG.error("Both pulp3-url and domain must be provided")
+            sys.exit(41)
+
+        # pulp-certificate provided as argument
+        if args.pulp_certificate:
```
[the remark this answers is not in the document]
rbikar: this should be with with new changes, re-review

src/pubtools/_pulp/services/pulp3.py
```diff
@@ -0,0 +1,109 @@
...
+        return pulplib.Pulp3Client(
+            args.pulp3_url, domain=args.domain, auth=auth, cert=cert
+        )
+
+    def __exit__(self, *exc_details):
+        # Note: The pulp3_client is an async context manager and must be
+        # properly exited in async code (e.g., using 'async with').
+        # We cannot call __aexit__ here as this is a synchronous method.
+        # Callers are responsible for managing the async context.
+        super(Pulp3ClientService, self).__exit__(*exc_details)
```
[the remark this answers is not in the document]
rbikar: this should be with with new changes, re-review

src/pubtools/_pulp/services/pulp3.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def get_pulp3_credentials(self):
+        out = None, (None, None)
+        if self._service_args.pulp3_user:
+            out = "basic", (
+                self._service_args.pulp3_user,
+                self._service_args.pulp3_password,
```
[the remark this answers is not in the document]
rbikar: this should be with with new changes, re-review

src/pubtools/_pulp/tasks/push/phase/sync.py
```diff
@@ -0,0 +1,190 @@
...
+            repo_name, pulp_labels=pulp_labels
+        )
+        task = await self.pulp3_client.create_distribution(
+            repo["pulp_href"],
+            name=distr_name,
+            base_path=distr_name,
+        )
+
+        await self.pulp3_client.poll_task(task)
+        distr = await self.pulp3_client.get_distribution(distr_name)
```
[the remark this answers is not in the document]
rbikar: this should be with with new changes, re-review

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py38,static,docs
+envlist = py310,static,docs
```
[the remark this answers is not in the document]
rbikar: won't fix, as we may need to keep some compatibility with older python, but it's out of scope

src/pubtools/_pulp/services/pulp3.py
rbikar: `auth` var fixed, the password isn't deemed as error in old pulp client so probably we can keep this as warning as well.

src/pubtools/_pulp/services/pulp3.py
rbikar: as mentioned above, the password isn't deemed as error in old pulp client so probably we can keep this as warning as well.

src/pubtools/_pulp/tasks/push/command.py
rbikar: @crungehottman 
Allowing all unit types there.
For Konflux source case:
- RPMs has to be synced in phase otherwise Upload will fail as src points to pulp href.
- Handling other types has not been decided yet.

src/pubtools/_pulp/tasks/push/phase/sync.py
rbikar: @crungehottman 
actually, I let Update phase to update this.

src/pubtools/_pulp/services/pulp3.py
rbikar: @crungehottman 
final solution is with error raised, as Pulp3Client with empty password (None) would raise error as well.

src/pubtools/_pulp/services/pulp3.py
rbikar: ...as commented elsewhere, I made the missing password as error.

rbikar:
@crungehottman 

I added a few minor fixes and tests for 100% coverage. 
Not fixing the static problems, looks unrelated.
Also the signing key setup for sync is not a problem. Destination repos by default require signed pkgs and allow specific keys + we enforce the same (at least requirement for signed pkg, not specific signature) for tmp repos for sync. Can be overridden by `--allow-unsigned` for testing purposes.

src/pubtools/_pulp/tasks/push/phase/upload.py
```diff
@@ -5,6 +5,8 @@
 from . import constants
 from ..items import State
 
+from pubtools.pulplib import YumRepository
```
crungehottman: Unused import, this seems to be the cause of the static analysis failure

src/pubtools/_pulp/tasks/push/phase/sync.py
```diff
@@ -0,0 +1,190 @@
...
+                ]:
+                    synced += 1
+                    self.put_output(item)
+                    continue
+
+                items_to_add.add(item)
+                syncing += 1
+
+            if items_to_add:
+                repo, dist_url = await self._create_ext_repo()
```
crungehottman: Would it be useful to generate the `rand_str` once in `_run`, and provide it to both `_create_ext_repo()` and `_create_tmp_repo()`  so that both repos share one UUID per sync run (instead of calling `rand_str = str(uuid.uuid4())` in both `_create_ext_repo()` and `_create_tmp_repo()`, resulting in two separate UUIDs for one sync run)? Using one UUID per sync run might be helpful for debugging purposes. (This could also be filed as a follow-up Jira, if that is preferred)

rbikar: yeah, sounds good.

crungehottman (approved):

